PR #67, opened 2017-01-03
fixed wrong path in opencl compiler call (opencl-rocm-dev 1.2.0)

psteinb:
based on `4.6.0-kfd-compute-rocm-rel-1.4-16`


psteinb:
any thoughts on this?

PR #110, opened 2017-05-03
Update README.md

briansp2020:
Fix the command to install opencl.


briansp2020:
I looks like it is now fixed.

PR #489, opened 2018-08-02
Update README.md

zhang2amd:
Added a comment to update the key if signature verification failed. Also updated key file hash since the key has extended expiration date.


sunway513:
LGTM

PR #529, opened 2018-09-14
Update README.md

ChristinaElder:


README.md
```diff
@@ -4,9 +4,7 @@ The ROCm Platform brings a rich foundation to advanced computing by seamlessly
 
 #### Supported CPUs
 
-Starting with ROCm 1.8, we have relaxed the requirements for PCIe Atomics on Vega 10 (GFX9) GPUs, and we have similarly opened up more options for number of PCIe lanes. With this release, these GFX9 GPUs can support CPUs without PCIe Atomics and, for example, run on PCIe  Gen2 x1 lanes. To enable this option, please set the environment variable `HSA_ENABLE_SDMA=0`.
-
-Currently, our GFX8 GPUs (Fiji & Polaris family) still need to use PCIe Gen 3 and PCIe Atomics, but are looking at relaxing this in a future release, once we have fully tested firmware.
+Starting with ROCm 1.8, we have relaxed the requirements for PCIe Atomics on Vega 10 (GFX9) GPUs, and we have similarly opened up more options for number of PCIe lanes. With this release, these GFX9 GPUs can support CPUs without PCIe Atomics and, for example, run on PCIe  Gen2 x1 lanes. To enable this option, please set the environment variable `HSA_ENABLE_SDMA=0`.  This is not supported on GPUs below GFX9, i.e. GFX8 cards in Fiji and Polaris families.
```
icarus-sparry: Should this be 1.9?

jlgreathouse: It should be 1.8, since that's when we relaxed the requirement. That said, I think the current description is incorrect -- this is not supported on GFX8 GPUs, but GFX7 (Hawaii) does not require PCIe atomics.

README.md
```diff
@@ -47,6 +45,57 @@ from the list provided above for compatibility purposes.
...
+
+
+#### New distribution support 
+
+* Binary package support for Ubuntu 18.04
+
+#### ROCm 1.9 is ABI compatible with KFD in upstream Linux kernels. 
+Upstream Linux kernels support the following GPUs in these releases:
+4.17: Fiji, Polaris 10, Polaris 11
+4.18: Vega10
```
jlgreathouse: Does 4.18 also support Fiji, Polaris 10, and Polaris 11?

README.md
```diff
@@ -47,6 +45,57 @@ from the list provided above for compatibility purposes.
...
+
+Some ROCm features are not available in the upstream KFD:
+* Less system memory available to ROCm applications
```
jlgreathouse: Our ROCm "feature" is that we offer less memory available? Or is this a problem that is *fixed* in ROCm compared to upstream?

README.md
fxkamd: Yes. What I was trying to say is, they were added in 4.17 and are supported going forward.

README.md
fxkamd: Right. This is inconsistent. I guess the missing feature is "More system memory available ..."

README.md
jlgreathouse: OK. Rather than resubmit the whole change, I'll mark this one as resolved and update it after the whole change is pushed.

PR #569, opened 2018-10-03
Updated doc on OS support

sunway513:
This PR specifies the ROCm recommended Ubuntu kernel versions.
And advise users to remove ROCm packages if need to upgrade the CentOS versions. There are known DKMS limitations can cause the system fail to upgrade if rock-dkms modules were installed.


icarus-sparry:
I made the typical rooky mistake and thought "This is a tiny change so I can use the web based editor to fix it"

PR #643, opened 2018-12-21
README and repo overhaul for the ROCm 2.0 release

jlgreathouse:
Community feedback has pointed out a number of confusing,
oudated, or missing sections in our ROCm README file. For example,
we do not describe what our ROCm package structure is, or how the
packages and meta-packages fit together. This can make it confusing
for users who do not want to just install rocm-dkms and move on.

Our repo manifest (default.xml) is severely out of date. It is
missing almost all of the current ROCm projects, and it always
pulls from the main development branch. This means we do not have
a pinned manifest that allows you to pull the code from a
particular ROCm reelease. Manifest updated, and the section of the
README discussing it is majorly overhauled (including links for
information/scripts about building the code after downloading it).

Rather than continually grow our version history in the main
README page, this splits off old version information into its own
file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 ## Are You Ready to ROCK?
-The ROCm Platform brings a rich foundation to advanced computing by seamlessly
-integrating the CPU and GPU with the goal of solving real-world problems.
-This software enables the high-performance operation of AMD GPUs for computationally-oriented tasks in the Linux operating system.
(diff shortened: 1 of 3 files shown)


jlgreathouse:
I still need to get final verification on the Polaris 12 support, so if anyone reads this PR -- note that at this exact moment, such a statement is not AMD-official.

Issue #659, opened 2019-01-03
fresh Ubuntu install without GPU not finding OpenCL platform


Noerr:
I tried to do a basic setup to test ROCm for OpenCL for CPU-only (initially).  clGetPlatformIDs is returning -1001.  Can anyone confirm these steps?

**Test setup:**
Fresh install of Ubuntu 18.04 **in a virtual machine**, running on Intel i7 v4 CPU with 4 cores, no discrete GPU.

**Steps:**

```
$ sudo apt update
$ sudo apt install rocm-dkms
```
 Succes. Reboot.

```
$ /opt/rocm/bin/rocminfo
hsa api call failure at line 900, file: /home/jenkins/jenkins-root/workspace/compute-rocm-rel-2.0/rocminfo/rocminfo.cc. Call retured 4104
```

```
$ /opt/rocm/opencl/bin/x86_64/clinfo
ERROR: clGetPlatformIDs(-1001)
```

`$ strace /opt/rocm/opencl/bin/x86_64/clinfo`
Shows that **libamdoclcl64.so** is not found in a couple dozen paths right before exit 1.

Did I miss something?  Other packages are mentioned on the install guide.

`$ sudo apt install rock-dkms`
"already the newest version" 0 upgraded, 0 newly installed.

`$ sudo apt install rocm-opencl-dev`
"already the newest version" 0 upgraded, 0 newly installed.

`$ sudo apt-get install dkms rock-dkms rocm-opencl`
for each ... "already the newest version" 0 upgraded, 0 newly installed.

I'm not sure what else to try.  Where should libamdoclcl64.so be located?   Would it be installed if I don't have a supported GPU?

Noerr:
Perhaps you can expand on the first sentence more explicitly please.  Are you saying there's no CPU target support in the whole ROC-whatever ecosystem?
I came to ROCm because I understood AMDAPPSDK was discontinued but redirected to this project.

jlgreathouse:
Correct, the ROCm OpenCL runtime does not support CPU targets. As far as I know, the OpenCL runtimes in the Linux/amdgpu-pro and [Windows](https://community.amd.com/message/2878639) software stacks also no longer support CPU targets, but I don't actively work in those areas so I can't verify that at this time.

Noerr:
Ok, thank you for the clarification.  I submitted a pull request to enhance the README. I think language like this would have saved me several hours down a dead-end path.
[...]

PR #661, opened 2019-01-03
enhancing documentation clarity with regards to CPU support

Noerr:
Addressing some language that got me confused, and related to [issue 659](https://github.com/RadeonOpenCompute/ROCm/issues/659).


Bengt:
I do agree that one might get the impression from the documentation that CPUs are supported as client devices. This makes it much clearer and is a step in the right direction in my view.

saadrahim:
stale PR

PR #698, opened 2019-02-06
Update tag for release 2.1

zhang2amd:


jlgreathouse:
LGTM, but we should probably wait to merge until all of the underlying packages make their proper roc-2.1.0 tags. At the moment, e.g., ROCR and rocprofiler haven't tagged roc-2.1.0 yet, so that would break a `repo sync`. Thoughts?

zhang2amd:
Agree. Evgeny is asking me the hash for his project. Probably only ROCR-Runtime remaining.

Thanks,
Zhongyu

PR #770, opened 2019-04-15
Fix llvm and lld 2.3 tags for opencl and hcc

kzhuravl:


searlmc1:
LGTM

nelsonc-amd:
Tested repo sync, seems fine.

PR #899, opened 2019-10-03
Use a newer OpenCL-ICD-Loader

acowley:
As of the 2.8 release, the ROCm-OpenCL-Runtime package no longer builds with the referenced revision of the Khronos OpenCL-ICD-Loader. This [commit to the ROCm-OpenCL-Runtime](https://github.com/RadeonOpenCompute/ROCm-OpenCL-Runtime/commit/2c8b286a597f055b9d0fad2d78e8193845cec772) makes it compatible only with ICD loaders at least as new as [this commit to the loader repository](https://github.com/KhronosGroup/OpenCL-ICD-Loader/commit/7d2584c4b8b1ac186f67143aaa2f846184709a9e). I have used the most recent ICD loader revision in this PR. cc @kzhuravl


saadrahim:
stale PR

PR #921, opened 2019-10-25
Update package structure.

zhang2amd:


README.md
```diff
@@ -311,18 +311,22 @@ The chain of software installed by these meta-packages is illustrated below
        |--hsa-amd-aqlprofile
        |--comgr
        |--rocr_debug_agent
+       |--rocprofiler-dev
```
icarus-sparry: Would it make sense to have these ordered alphabetically?

zhang2amd: I thought about it but didn't want to change too much so just make packages in groups in rocm-libs. But now you also mentioned this so I made it alphabetical except rocm-utils.

PR #986, opened 2019-12-29
Update instruction for Debian-based with Upstream Kernel Drivers

kolmar-kafran:
There is a typo error on Using Debian-based ROCm with Upstream Kernel Drivers. The pipe was missing.


saadrahim:
stale PR

Issue #930, opened 2019-11-07
rocm installation issue in my pc


kabish01:
hsa api call failure at: /data/jenkins_workspace/compute-rocm-rel-2.9/rocminfo/rocminfo.cc:1102
Call returned HSA_STATUS_ERROR_OUT_OF_RESOURCES: The runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events.

Burstaholic:
What hardware are you using? This happens to me on my Vega M laptop (i7-8705G).

kabish01:
I am using Ryzen 5 3550h CPU with Vega 8 integrated graphics and Radeon RX560X dedicated GPU.

bugparty:
amd 1800x + amd470 and centos8.3 failed too

Shreyashwaghe:
Any luck with the recent rocm 4.1 or 4.0 with your rx560x card, i am facing few problems too

nartmada:
Original ticket is more than a year old and the person that opened the ticket has not responded to the latest request.  If this is still an issue, please file a new ticket and we will investigate.  Thanks!

Issue #923, opened 2019-10-26
ROCm not correctly installed


creekbed:
Hi all. First time posting. I'm new to Linux so I'm endeavoring to tackle the hail of problems I'm getting. I'm aware I may be or have been doing something wrong, or extremely dumb, so that's why I'm here. I wanted to see if ROCm was successfully installed (below you'll see what I did to do so) by running the following commands expecting my GPU to be listed, but I get the following:

```
root@slack:~# /opt/rocm/bin/rocminfo 
ROCk module is loaded
root is member of video group
hsa api call failure at: /data/jenkins_workspace/compute-rocm-rel-2.9/rocminfo/rocminfo.cc:1102
Call returned HSA_STATUS_ERROR_OUT_OF_RESOURCES: The runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events.
`
and (I think there's no problem with this one, although at the very bottom there's something wrong):
root@slack:~# /opt/rocm/opencl/bin/x86_64/clinfo 
Number of platforms:				 2
  Platform Profile:				 FULL_PROFILE
  Platform Version:				 OpenCL 1.2 pocl 1.3 None+Asserts, LLVM 8.0.1, SLEEF, DISTRO, POCL_DEBUG
  Platform Name:				 Portable Computing Language
  Platform Vendor:				 The pocl project
  Platform Extensions:				 cl_khr_icd
  Platform Profile:				 FULL_PROFILE
  Platform Version:				 OpenCL 2.1 AMD-APP (2982.0)
  Platform Name:				 AMD Accelerated Parallel Processing
  Platform Vendor:				 Advanced Micro Devices, Inc.
  Platform Extensions:				 cl_khr_icd cl_amd_event_callback cl_amd_offline_devices 


  Platform Name:				 Portable Computing Language
Number of devices:				 1
  Device Type:					 CL_DEVICE_TYPE_CPU
  Vendor ID:					 6c636f70h
  Max compute units:				 2
  Max work items dimensions:			 3
    Max work items[0]:				 4096
    Max work items[1]:				 4096
    Max work items[2]:				 4096
  Max work group size:				 4096
  Preferred vector width char:			 16
  Preferred vector width short:			 16
  Preferred vector width int:			 8
  Preferred vector width long:			 4
  Preferred vector width float:			 8
  Preferred vector width double:		 4
  Native vector width char:			 16
  Native vector width short:			 16
  Native vector width int:			 8
[...]

Issue #889, opened 2019-09-17
can‘T


Bitllion:

Bitllion:
/opt/rocm/bin/rocminfo
/opt/rocm/opencl/bin/x86_64/clinfo

ROCk module is loaded
root is member of video group
hsa api call failure at: /data/jenkins_workspace/compute-rocm-rel-2.7/rocminfo/rocminfo.cc:1102
Call returned HSA_STATUS_ERROR_OUT_OF_RESOURCES: The runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events.
Number of platforms:                             1
  Platform Profile:                              FULL_PROFILE
  Platform Version:                              OpenCL 2.1 AMD-APP (2949.0)
  Platform Name:                                 AMD Accelerated Parallel Processing
  Platform Vendor:                               Advanced Micro Devices, Inc.
  Platform Extensions:                           cl_khr_icd cl_amd_event_callback cl_amd_offline_devices


  Platform Name:                                 AMD Accelerated Parallel Processing
ERROR: clGetDeviceIDs(-1)

ITfirewall:
Hey Bitllion, 

 Whats your OS and Gpu???

jppyykm:
I have a similar problem. I am running Ububtu and my gpu is RX 480. 
```
hsa api call failure at: /data/jenkins_workspace/compute-rocm-rel-2.7/rocminfo/rocminfo.cc:1102
Call returned HSA_STATUS_ERROR_OUT_OF_RESOURCES: The runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events.
```
https://rocm.github.io/ROCmInstall.html  <-- Following installation 
 
Installing this thing has been really, really annoying. I broke my Mint installation a few times tinkering to make it run there. I got tired and installed Ubuntu so it is supported. Now this.

ITfirewall:
Hi jppyykm,

Can you try again with an older version of ROCm driver??? ( try ROCm 2.6, 2.5,....) from repository.
I have a different problem with ubuntu 16.04 + ROCm 2.7.1, this is causing a problem with OpenCL.

jppyykm:
Hi! I am having all sorts of trouble. I am working on it in sprints.. getting this thing to work is urgh,,

b4551k5:
I got the same issue. It says:

[...]

PR #1005, opened 2020-01-24
Update README.md

sileht:
I tried it on a Debian buster with a recent kernel (5.4.14).

Debian kernel configuration does not have all required options by default, so I added a command to check the 3 options I have to add.

Also on recent debian DRI devices are created with render group not video group (video is still need for /dev/fkd).

    ls -la /dev/dri/renderD129
    crw-rw---- 1 root render 226, 129 Jan 24 09:45 /dev/dri/renderD129

This may fix this following issues that have same kind of error I got:
https://github.com/RadeonOpenCompute/ROCm/issues/930
https://github.com/RadeonOpenCompute/ROCm/issues/923
https://github.com/RadeonOpenCompute/ROCm/issues/889

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,12 +271,13 @@ To install from a Debian Repository:
 5. To add your user to the video group, use the following command for the sudo password:
 	
 		sudo usermod -a -G video $LOGNAME
+		sudo usermod -a -G render $LOGNAME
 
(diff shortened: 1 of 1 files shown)


saadrahim:
stale PR

PR #1048, opened 2020-03-16
Fewer worries transitioning from AMDGPU driver

smoe:
Users coming from the official AMDGPU driver on Ubuntu should be informed that they are not losing out when the amdgpu-dkms package is removed. rock-dkms and amdgpu-dkms cannot both be installed at the same time.

I moved the "apt update" to where the repository info was changed since this is where in my mind it belongs.


saadrahim:
stale PR

PR #1049, opened 2020-03-17
improvements to SLES installation instructions README.md

clavinet:
some lines were indented wrong, leading to broken formatting


clavinet:
I also changed it to add the repo as https:// since loading binaries without gpg signature checking and without even TLS is rather insecure. This should probably added to other distros too, but I don't know if they support it.

saadrahim:
stale PR

PR #1154, opened 2020-06-19
Roc 3.5.1 : Manifest Files

adilad:


codecor-cn:
so many errors on this time update, /opt/rocm has gone away and many library not found

PR #1550, opened 2021-08-07
Fix missing links in rocprof docs

arghdos:


arghdos:
@Rmalavally -- can you take a look? Cleaned up the rocprof section a bit.

PR #1588, opened 2021-10-12
Markdownify list of supported operating systems

Bengt:
This uses GitHub-flavored Markdown to create the table, which lists the supported operating systems. When compared to the former image, this helps with screen readers, search engine optimization and also with maintenance of this code base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,15 @@ It also covers known issues in this release.
 
 The AMD ROCm platform supports the following operating systems:
 
-![Screenshot](https://github.com/RadeonOpenCompute/ROCm/blob/master/images/SuppEnv.PNG)
+| Distribution   |        Kernel       |
+| -------------- | ------------------- |
+| SLES15 SP3     | 5.3.18-24.49        |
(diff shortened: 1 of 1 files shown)


icarus-sparry:
Good improvement.

keryell:
Thanks!

saadrahim:
Closing this PR as documentation is no longer contained in this repository.

saadrahim:
At the moment documentation is not maintained in this repository or on github.com. A mechanism for public feedback for documentation is planned.

Bengt:
I see. Thanks for the information. How is the process of providing feedback like this to the documentation if there is no mechanism currently in place?

saadrahim:
If you take a look at https://docs.amd.com/bundle/ROCm-Release-Notes-v5.2/page/About_This_Document.html, there is a small feedback icon on the right side of the page. That is the current process to submit feedback.

saadrahim:
As the change entered the documentation already, I will close this PR.

Bengt:
Awesome. Thanks for the guidance, @saadrahim!

PR #1737, opened 2022-05-16
Adds hipRAND to default.xml

dipietrantonio:
Missing hipRAND line in the default.xml file makes such that hipRAND is not downloaded when using  the`repo` command. This commit adds the missing line.

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -54,6 +54,7 @@ fetch="https://github.com/KhronosGroup/" />
     <project name="rocFFT" remote="rocm-swplat" />
     <project name="hipFFT" remote="rocm-swplat" />
     <project name="rocRAND" remote="rocm-swplat" />
+    <project name="hipRAND" remote="rocm-swplat" />
     <project name="rocSPARSE" remote="rocm-swplat" />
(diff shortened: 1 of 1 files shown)


saadrahim:
hipRAND is missing in default.xml due to a recent change. The hipRAND library used to be included in the rocRAND repository. This was confusing for end users and build systems. During a transition period, hipRAND has moved to a separate repository and is available as a submodule in rocRAND. The last step to remove the submodule of hipRAND in rocRAND is not complete. @lawruble13 can you please follow up on this issue?

Merging this PR without making the necesasary internal build process fixes may not be the best longterm fix. If this is needed ASAP, I can reconsider.

@dipietrantonio We have noticed you are a power user of ROCm. Feel free to reach out to me at saad . rahim @amd.com (less the spaces) for any concerns. I am gathering feedback on usability and would appreciate a discussion.

dipietrantonio:
@saadrahim Thanks for the explanation, this is not urgent so happy to wait you fix this internally. I would be happy to provide feedback and establish a communication channel! I will be reaching out soon as possible, I appreciate it.
Regards

PR #1823, opened 2022-10-06
Adding stakeholders to CODEOWNERS file

saadrahim:


Rmalavally (changes requested):
Could we please have a general ROCm repository that we can all access? Thanks so much.

saadrahim:
All github repositories will be branch protected. There is no free for all access. Once we have enough code reviewers, administrator privileges will be removed as well.

PR #1945, opened 2023-03-14
Rtd alexv feedback

AlexVlx:
This holds an assorted set of additional cleanups, primarily focused on:

- keep line length under 80-characters, except under extreme duress (this allows side-by-side views of source & previewed / rendered text);
- the source `.md` should be readable in unrendered form i.e. one should be able to read the source file linearly, so e.g. tables should be readable even if Markdown parsers don't really care about things like aligning pipes
- abuses of `\` as introductor of special characters are removed, we have inline codeblocks for this.

Some bits are noisy due to the switch to the develop branch here and `git` being funky (i.e. there's no substantive, visible change that the diff spots).


saadrahim (approved):

saadrahim:
Was this formatted with markdown lint?

PR #1963, opened 2023-03-16
Pin rocm-docs-core to a specific commit and update requirements

Maetveis:
Pinning `rocm-docs-core` to a specific version makes the requirements entirely reproducible and keeps a record of the version the docs can be built with.
This also makes sure CI can check when some new features are being used, but the version was not updated.


saadrahim:
Due to the rapid rate of development, we will not pin rocm-docs-core until the repository stabilizes. We will tag rocm-docs-core to allow for easier pinning.

saadrahim (changes requested):
Please do not merge now.

PR #1966, opened 2023-03-17
Add minimal documentation for ROCr

AlexVlx:
As per the title, this adds the minimal amount of documentation for ROCr:

- A general discussion of most relevant aspects (building, philosophy, interfaces etc.);
- The autogenerated API documentation, in `.pdf` form for now (casting it into either editable `.md` or sphinx consumable xml is left as a follow-up, pending a few adjustments in the ROCr repo itself).

Please consider if this is sufficient for this stage of the docs project.


docs/reference/rocr/rocr_api.pdf
saadrahim, line 1: This one will have to wait for a ROCr repository based documentation. No shortcuts here as no PDF based documentation should be used (except in dire circumstances). I will add this to @samjwu 's queue. It is best to do the conversion at the doxygen level.

saadrahim:
Closing in favor of ROCR-Runtime#156

PR #1974, opened 2023-03-19
Demonstrating figure and table caption standard

saadrahim:


saadrahim:
@Ehud-Sharlin please see the syntax here for adding figure and table captions.

PR #1976, opened 2023-03-20
Initial MI250 Guide

### LINKED ISSUE #1928: Article for introduction to MI250
Please add copy the MI250 architecture article from https://docs.amd.com/bundle/AMD-Instinct-MI250-High-Performance-Computing-and-Tuning-Guide-v5.3/page/AMD_Instinct_Hardware.html and add to the GPU architectures section.

MathiasMagnus:
Closes #1928


saadrahim (changes requested):
Please update figures and table numbering to automated.

MathiasMagnus:
Numbered references don't work. Even if I set `numFig = True` in `conf.py`, it won't assign numbers to the images/tables. If we add a name so we can reference the image, it won't get a number, if we don't assign a name, it get's a number but can't be refereced from text.

MathiasMagnus:
@lawruble13 it doesn't work, no matter what I do. Here, it was likely a typo. `conf.py` ending with:
```
for sphinx_var in ROCmDocs.SPHINX_VARS:
    globals()[sphinx_var] = getattr(docs_core, sphinx_var)

numfig = True
```
And `mi250.md` containing:
``````
{numref}`mi250-gcd` shows the components ...

```{figure} ../../data/reference/gpu_arch/image.001.png
:scale: 100%
:align: center
:name: mi250-gcd

Structure of a single GCD in the AMD Instinct MI250 accelerator.
```
``````
And I get the following warning during build:
```
C:\Users\mate\Source\Repos\ROCm\docs\reference\gpu_arch\mi250.md:15: WARNING: Failed to create a cross reference. Any number is not assigned: mi250-gcd
```
And indeed, the rendered text is not a link, just "`mi250-gcd` shows the components"... This supposedly _should_ work, as far as [MyST Cheat Sheet](https://jupyterbook.org/en/stable/reference/cheatsheet.html#figures-and-images) or the [reference docs](https://jupyterbook.org/en/stable/content/references.html) go. It doesn't matter what order the numref and the figure definition is, I tried reversing the order, but that has no effect.

**Comment by MathiasMagnus** [comment]
I spent the better part of Yesterday looking trying to figure this out, looking at docs, guides, the MyST reference, the source code of rocm-docs-core but I can't find where it goes wrong.

File: docs/reference/gpu_arch.md [hunk 1]
```diff
@@ -17,3 +17,17 @@
...
+## Architecture Guides
+
+:::::{grid} 1 1 1 1
+:gutter: 1
+
+:::{grid-item-card} AMD Instinct MI200
+This chapter briefly reviews hardware aspects of the AMD Instinct MI250 accelerators and the CDNA™ 2 architecture that is the foundation of these GPUs.
+
+- [Instruction Set Architecture](https://www.amd.com/system/files/TechDocs/instinct-mi200-cdna2-instruction-set-architecture.pdf)
+- [Whitepaper](https://www.amd.com/system/files/documents/amd-cdna2-white-paper.pdf)
```
AlexVlx, line 30: Considering recent... events, I wonder if we should consider including these locally, in the repository, as non-code/non-documentation binary artifacts (i.e. under `/docs/data/foo`), just like the images? Opinions, @saadrahim ?

lawruble13:
Can I request that you try using [Markdown figures](https://jupyterbook.org/en/stable/content/figures.html#markdown-figures)? You can reference #1980, which we have validated as working.

docs/reference/gpu_arch.md
saadrahim: I am not sure if we want it locally. If the binary artifacts are already stored on an AMD site, isn't it better to host it there? I understand links maybe broken by third parties, I am not sure if growing this repo size is the correct response.

MathiasMagnus:
@lawruble13 I moved all references to numref. The trick to getting the image references working was to convert _all_ figures, and then the `Any number is not assigned` warning went away.

Note that the only `list-table` entry gets it's numbering wrong (it starts with Table 2), but at least it's automated. I'd consider that to be a minor issue and eligible for deferral.

PR #1980, opened 2023-03-21
Figure update to figure-md

saadrahim:


Rmalavally (commented):
@saadrahim Have we established file names for images? If we are referencing images by name rather than number, it can be prone to error. Especially with multiple versions and images. What is the process to ensure developers have used the correct image? How would documentation know for review purposes? Assume all images will be stored in the images folder.

saadrahim:
> @saadrahim Have we established file names for images? If we are referencing images by name rather than number, it can be prone to error. Especially with multiple versions and images. What is the process to ensure developers have used the correct image? How would documentation know for review purposes? Assume all images will be stored in the images folder.

Yes, images need to use snake_case and not numbering. Some of the older image names need changing.

Rmalavally:
Thanks, Saad. Snake_case is fine for files. 

We use camelCase for method names and variables and underscores for functions. Will the integrated linter bypass these cases? Does it require additional configuration? Is this documented in the contribution guide?

Rmalavally:
Thanks, Saad. Sorry, I was not clear in my question. Will the linter identify the CamelCase as a typo? Can it be customized to bypass CamelCases? If not, we must manually correct all instances, which will be many.

saadrahim:
There is no linting planned for camelcase checks at moment. If we do enable it, it will be for file names only.

PR #1984, opened 2023-03-22
ROCmCC & Win Install: Table & Figures Cleanup

### LINKED ISSUE #1955: Formatting for some tables looks odd
image

Was discussing with @AlexVlx

Mainly this seems to result from putting the table header in a single column (so you get fixed width + wrap).  Not clear what the best solution is here, perhaps splitting it to be e.g., "Table 1" in the top row (to keep it from wrapping / looking out of place), and put a 'caption' below in some other format?

Naraenda:
Resolves #1955

Scope
- Replaces table captions and headers with MyST-style captions.
- Replaces some tables with the `:::{option}`-directive.
- Uses MyST-style figures for the Windows SDK Installation guide.
- Uses numref whenever possible.

Note for reviewer: I did not touch the clang options table, apart from wrapping it into a table directive. Is this table auto-generated? If so, how? Then we can use the same `:::{option}` directives for the arguments ^^

I.e.
```
:::{option} -dsym-dir <dir>
**Unsupported!**
Outputs dSYMs (if any) to the specified directory.
:::
```

### DIFF
--- a/docs/hip_sdk_install_win/hip_sdk_install_win.md
+++ b/docs/hip_sdk_install_win/hip_sdk_install_win.md
@@ -11,16 +11,18 @@ TODO: provide link to supported GPU guide.
 
 ## SDK Installation
 
-Installation options are listed in Table 1.
+Installation options are listed in {numref}`installation-components`.
 
-| **Table 1. Components for Installation** |||                                                                                                                                                                                                      |
-|:------------------------:|:----------------:|:------------------:|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|
+:::{table} Components for Installation
+:name: installation-components
 | **HIP Components**       | **Install Type** | **Display Driver** | **Install Options**                                                                                                                                                            |
+|:------------------------:|:----------------:|:------------------:|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|
 | **HIP SDK Core**         | **Full**         | Adrenalin 22.40    | **Full:** Provides all AMD Software features and controls for gaming, recording, streaming, and tweaking your performance on your graphics hardware.                           |
 | **HIP Libraries**        | **Full**         |                    | **Minimal:** Provides only the basic controls for AMD Software features and does not include advanced features such as performance tweaking or recording and capturing content.|
 | **HIP Runtime Compiler** | **Full**         |                    | **Driver Only:** Provides no user interface for AMD Software features.                                                                                                         |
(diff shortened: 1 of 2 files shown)


docs/hip_sdk_install_win/hip_sdk_install_win.md
```diff
@@ -72,48 +80,62 @@ to [Installing Components](#installing-components).
...
-| ![HIP Ray Tracing](../data/hip_sdk_install_win/HIP-Ray-Tracing.png) |
-|:-------------------------------------------------------------------:|
-| **Figure 8. HIP Ray Tracing**                                       |
-
-| ![BitCode Profiler](../data/hip_sdk_install_win/BitCode-Profiler.png) |
-|:---------------------------------------------------------------------:|
-| **Figure 9. BitCode Profiler**                                        |
+shown in {numref}`hip-sdk-core`.
+
+:::{figure} ../data/hip_sdk_install_win/HIP-SDK-Core.png
```
saadrahim, line 110: We should switch to figure-md syntax for better markdown rendering.

docs/reference/rocmcc/rocmcc.md
```diff
@@ -6,30 +6,32 @@ ROCmCC is a Clang/LLVM-based compiler. It is optimized for high-performance
...
+ROCmCC is made available via two packages: `rocm-llvm` and `rocm-llvm-alt`.
+The differences are listed in [the table below](rocm-llvm-vs-alt).
+
+:::{list-table} Differences between `rocm-llvm` and `rocm-llvm-alt`
+:header-rows: 1
+:name: rocm-llvm-vs-alt
+
+* - **rocm-llvm**
+  - **rocm-llvm-alt**
+* - Installed by default with ROCm™
```
saadrahim, line 18: Let's not switch to list tables. We can discuss tomorrow but end goal is to eventually comply with GitHub Markdown

saadrahim (changes requested):
Let's discuss this tomorrow to go over a few concerns and future direction.

docs/reference/rocmcc/rocmcc.md
Naraenda: Reverted in [ab9a035](https://github.com/RadeonOpenCompute/ROCm/pull/1984/commits/ab9a03547b07efe20e17f66c5f332a25e2e43443)

docs/hip_sdk_install_win/hip_sdk_install_win.md
Naraenda: The HTML output is the same and they can be used interchangeably. Do we still want to switch?

PR #2005, opened 2023-03-29
Enabling markdown lint on PRs

saadrahim:


samjwu (approved):

saadrahim:
Markdown lint is still going to fail but this way everyone knows about it. It will be required soon.

PR #2025, opened 2023-04-06
Added reference to AMD lab notes

jychang48:
It would be great to also provide readers a pointer to the public AMD lab notes series

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -88,6 +88,7 @@ Examples
 ^^^
 
 - [ROCm Examples](https://github.com/amd/rocm-examples)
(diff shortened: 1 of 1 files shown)


saadrahim (approved):
Looks good to me. 

Let's have this as the start of a fruitful collaboration.

saadrahim:
@jychang48 can also provide the links to your blogs. Let's find a place to integrate those as well.

jychang48:
@saadrahim sure, let me discuss with the team to see how we can best make this happen

saadrahim:
> @saadrahim sure, let me discuss with the team to see how we can best make this happen

All I am thinking was to put in the links to the GPUOpen site.

PR #2037, opened 2023-04-12
Readme Cleanup

saadrahim:


Rmalavally (commented):
Do we have anyone from release management reviewing this description? We must be careful about generic descriptions of backward compatibility.

saadrahim:
> Do we have anyone from release management reviewing this description? We must be careful about generic descriptions of backward compatibility.

The versioning description has been dropped for now.

PR #2067, opened 2023-04-19
Update docker.md

doscherda:
add --security-opt seccomp=unconfined info

### DIFF
--- a/docs/deploy/docker.md
+++ b/docs/deploy/docker.md
@@ -51,6 +51,19 @@ like:
 docker run --device /dev/kfd --device /dev/dri/renderD128 --device /dev/dri/renderD130 <image>
 ```
 
+### Additional Options
+
+The performance of an application can vary depending on the assignment of GPUs and CPUs to the task. Typically, numactl is installed as part of many 
+HPC applications to provide GPU/CPU mappings. This Docker runtime option supports memory mapping and can improve performance
+
+```--security-opt seccomp=unconfined```
+  
+This option is recommened for Docker Containers running HPC applications.
+
+```shell
+docker run --device /dev/kfd --device /dev/dri --security-opt seccomp=unconfined ...
+```
+
 ## Docker images in the ROCm ecosystem
 
 ### Base images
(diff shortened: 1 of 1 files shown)


docs/deploy/docker.md
```diff
@@ -51,6 +51,19 @@ like:
...
 ```
 
+### Additional Options
+
+The performance of an application can vary depending on the assignment of GPUs and CPUs to the task. Typically, numactl is installed as part of many 
+HPC applications to provide GPU/CPU mappings. This Docker runtime option supports memory mapping and can improve performance
+
+```--security-opt seccomp=unconfined```
+  
+This option is recommened for Docker Containers running HPC applications.
```
saadrahim, line 61: Why is it recommended for HPC applications? It would be nice to know as this is impacts the end users security.

doscherda: Perhaps I am missing a ":"  

Would this be clearer -   This Docker runtime option supports memory mapping and can improve performance:

The line should describe the --security-opt option and explains why (performance) it is recommended. Please make any change needed to clarify that the security option is recommended because it allows for memory mapping and (therefore) improved performance.

saadrahim: Now it is getting more interesting. I would like to link an article on our memory model and why it needs the --security-opt. Something for the future.

Maetveis: Disabling seccomp filtering is a bad recommendation to make, it negates most security benefits docker gives you.

It would be much better to enable the syscall on top of docker defaults that is needed for better performance.

I couldn't find any mention of memory mapping that would be blocked (`mmap` and `mmap2` are allowed) by the default config. `mbind` is blocked though by default, which is for manipulating NUMA policies, maybe this is what you meant?
For this enabling the `CAP_SYS_NICE` capability is sufficient, but I would still mention where this might be needed (e.g. in NUMA aware applications that use the syscall or for using `numactl` in the container) instead of a blanket recommendation.

saadrahim: I understand that the option is important to add for performance. However, let me do a little due diligence to see if there is an alternate secure mechanism to get the performance improvement.

Naraenda:
Rebased & made CI pass.

saadrahim (approved):

PR #2079, opened 2023-04-24
Clean up the deployment related pages

Maetveis:
- Add an index page for the linux deployment submenu
- Remove deployment options that are not yet completed (i.e. spack, from source installation)
- remove the general deployment index page
- various cleanups and clarifications in the rest of the pages


saadrahim (changes requested):
Match git directory structure in end HTML

Maetveis:
Spelling check fails because the wordlist doesn't have OEM: https://github.com/RadeonOpenCompute/rocm-docs-core/pull/119 add's it in rocm-docs-core

PR #2085, opened 2023-04-26
Removing Windows Documentation

saadrahim:


iAvoe:
He's just removing the document, not the main program, but saadrahim didn't say anything regard this action, though

> why ???? we waited for windows release

radudiaconu0:
yeah... i dont know why he would remove it.

PR #2090, opened 2023-04-26
Set article info for pages

samjwu:


docs/conf.py
```diff
@@ -40,6 +40,40 @@
     "http://cs231n.stanford.edu/"
 ]
 
+html_output_directory = "_build/html"
+setting_all_article_info = True
+all_article_info_os = ["linux", "windows"]
+all_article_info_author = ""
+all_article_info_date = "May 1, 2023"
```
saadrahim, line 47: I was under the impression you were getting the date with  : :date: "{sub-ref}`today`" and time with : :read-time: "{sub-ref}`wordcount-minutes` min read"

From https://sphinx-design.readthedocs.io/en/sbt-theme/additional.html

samjwu: MyST substitutions require myst-parser version 1.0.0, but our myst-nb dependency requires myst-parser to be version 0.18 (https://github.com/executablebooks/MyST-NB/blob/master/pyproject.toml)

My workaround was to insert the HTML into the output, which has the added benefit of being independent of the source language (i.e.: RST or Markdown) and avoids having to replace an article info header if it already exists in the source file

saadrahim:
@samjwu how come the date is off?

samjwu:
> @samjwu how come the date is off?

it turns out Read the Docs uses git version 2.17.1, which does not support the %cs format option

https://github.com/RadeonOpenCompute/rocm-docs-core/pull/163 uses an older format that is supported

saadrahim (dismissed):

saadrahim:
Please change to PyPi package.

saadrahim:
I don't think realign reqs will work, don't we have to publish to PyPi?

docs/conf.py
```diff
@@ -42,7 +42,40 @@
...
+    {"file":"how_to/magma_install/magma_install", "os":["linux"]},
+    {"file":"how_to/pytorch_install/pytorch_install", "os":["linux"]},
+    {"file":"how_to/tensorflow_install/tensorflow_install", "os":["linux"]},
+    {"file":"examples/ai_ml_inferencing", "os":["linux"]},
+    {"file":"examples/inception_casestudy/inception_casestudy", "os":["linux"]},
+    
+    {"file":"understand/isv_deployment_win", "os":["windows"]},
+]
+
+docs_core = ROCmDocs("ROCm Docs 5.6.0 Alpha")
```
saadrahim, line 78: remove alpha

samjwu: Done

saadrahim (approved):

PR #2096, opened 2023-05-01
What is ROCm? - Overview

saadrahim:


docs/index.md
```diff
@@ -1,9 +1,27 @@
...
+- {doc}`/deploy/quick_start_windows`
+- {doc}`/deploy/docker`
 
-:::::{grid} 1 1 2 2
-:gutter: 1
+:::
+::::
+
+::::{grid-item}
+:::{dropdown} [What is ROCm?](rocm)
```
samjwu, line 17: It might be more intuitive to have this directly on the page (as in not in a grid), or as the first grid item (i.e.: before `Deploy ROCm`) if there are concerns about saving space.

saadrahim: Space isn't that much of a concern but you are right, it could be the first before deploy. I was thinking differently. I wanted What is ROCm to eventually be a gateway to how we do ROCm releases, how we version, explanations on userspace, kernel space, etc.

samjwu (dismissed):

samjwu (approved):

PR #2104, opened 2023-05-03
CMake HIP language support

### LINKED ISSUE #1957: Add HIP-Language support to "Using CMake" page
Main points are:

- Requires CMake >= 3.21
- CMAKE_HIP_COMPILER=clang++ (in place of CMAKE_CXX_COMPILER=hipcc)
- Suffix of HIP files is '.hip', or use `source_files_properties` + `LANGUAGE hip` (https://cmake.org/cmake/help/latest/command/set_source_files_properties.html)
- Brief discussion of hip-lang:: targets

I can help contribute here, @saadrahim

MathiasMagnus:
Closes #1957 

This PR intends to address the conversation points brought up in the linked issue, and adds the following changes:

- Simplifies the introduction paragraph which (IMHO) goes into too much detail.
  - Discussing details are moved to subsequent sections.
- The existing guide was Linux only. Windows-specifics were added on equal grounds.
  - Mentions on how to use CMake on Windows don't go into greater detail than the myriad of build definitions already public on GitHub.
- An attention drawing admonition was added to the _Compiling device code in C++ language mode_ section stating that while it is described for completeness' sake, it is a deprecated workflow. Users not being bound to CMake version prior to 3.21 should not piggyback on top of CXX language support and should move to HIP language support.
  - I personally would encourage ecosystem libraries to follow suit at their earliest convenience.
- A section dedicated to CMake presets were added.
  - This is a feature of CMake I hope ROCm ecosystem libraries will embrace to both simplify CI scripts, help users repeat the very same CMake invocations as done in CI, and demonstrate how to register toolchain installations in a reusable fashion.
  - A note is added that using presets on Windows isn't as straightforward, because HIP support for CMake's MSBuild generator is still absent and current tooling support don't help users incorporate environment setup scripts easily into preset files. (Solutions exist, see [here](https://gitlab.kitware.com/cmake/cmake/-/issues/21619#note_1356844) for details on workarounds.)


docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
 Using CMake
-===========================
+***********
 
-Most components in ROCm support CMake 3.5 or higher out-of-the-box and do not
-require any special Find modules. A Find module is often used by downstream to
-find the files by guessing locations of files with platform-specific hints.
-Typically, the Find module is required when the upstream is not built with CMake
-or the package configuration files are not available.
+Most components in ROCm support CMake. Projects depnding on header-only or
```
arghdos, line 5: 'depnding'->'depending'

saadrahim:
@MathiasMagnus I wonder if we should convert to markdown to match the other files. Originally, this was in rst. No other reason for it to remain.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+In short, CMake supports finding dependencies in two ways:
+
+-  In Module mode, it consults a file ``Find<PackageName>.cmake`` which tries to
+   find the component in typical install locations and layouts. CMake ships a
+   few dozen such scripts, but users and projects may ship them as well.
+-  In Config mode, it locates a file named ``<packagename>-config.cmake`` or
+   ``<PackageName>Config.cmake`` which describes the installed component in all
+   regards needed to consume it.
+
+ROCm dominantly relies on Config mode, one notable exception being the Module
```
arghdos, line 30: predominantly?

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+driving the compilation of HIP programs on Nvidia runtimes. As such, when
```
arghdos, line 31: I feel as if the difference between these two would be more clear with a simple hello-world (e.g., do a printf from the GPU) for both the AMD and NV backends to see how the CMake would differ.  I'm happy to defer that to a subsequent issue though.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+``-D CMAKE_PREFIX_PATH=....`` . AMD packaged ROCm installs can typically be
+added to the config file search paths such as:
+
+-  Windows: ``-D CMAKE_PREFIX_PATH=${env:HIP_PATH}``
+
+-  Linux: ``-D CMAKE_PREFIX_PATH=/opt/rocm``
 
 ROCm provides the respective *config-file* packages, and this enables
 ``find_package`` to be used directly. ROCm does not require any Find module as
 the *config-file* packages are shipped with the upstream projects.
```
arghdos, line 49: What are "the upstream projects" in this context?

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+==================
 
-When dependencies are not found in standard locations such as */usr* or
-*/usr/local*, then the ``CMAKE_PREFIX_PATH`` variable can be set to the
-installation prefixes. This can be set to multiple locations with a semicolon
-separating the entries.
+ROCm componenents providing a C/C++ interface support being consumed using any
+C/C++ toolchain that CMake knows how to drive. ROCm also supports CMake's HIP
+language features, allowing users to program using the HIP single-source
+programming model. The HIP API without compiling GPU device code behaves as a
```
arghdos, line 60: When a program (or translation-unit) uses the HIP API without compiling any GPU device code, HIP can be treated in CMake as a simple C/C++ library.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+C/C++ library.
 
-There are two ways to set this variable:
+Using the HIP single-source programming model
+---------------------------------------------
 
--  Pass the flag when configuring with ``-DCMAKE_PREFIX_PATH=....`` This
-   approach is preferred when users install the components in custom locations.
+Source code written in the HIP dialect of C++ typically use the `.hip`
```
arghdos, line 66: Either:

Source code*s* written ... typically use

Or

Source code written ... typically use*s*

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+fragments), enabling the C++ language is also recommended.
 
--  Append the variable in the CMakeLists.txt file. This is useful if the
-   dependencies are found in a common location. For example, when the binaries
-   provided on `repo.radeon.com <http://repo.radeon.com>`_ are installed to
-   */opt/rocm*, you can add the following line to a CMakeLists.txt file
+::
 
-   ::
+    cmake_minimum_required(VERSION 3.21) # enable_language(HIP)
```
arghdos, line 74: I think the comment here is trying to say "you need 3.21 to enable the HIP language"?

If it's instead trying to show that HIP can also be enabled by `enable_language`, I would put the comment next to the `project` call saying "# or, the user can call `enable_language(HIP)` directly"

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+
+Which use the device compiler provided from the binary packages of
+`ROCm HIP SDK <https://www.amd.com/en/graphics/servers-solutions-rocm>`_ and
+`repo.radeon.com <https://repo.radeon.com>`_ respectively.
 
+When using the CXX language support to compile HIP device code, selecting the
+target GPU architectures is done via setting the ``GPU_TARGETS`` variable.
+``CMAKE_HIP_ARCHITECTURES`` only exists when the HIP language is enabled. By
+default, this is set to some subset of the currently supported architectures of
+AMD ROCm. It can be set for eg. like ``-D GPU_TARGETS="gfx1032;gfx1035"``.
```
arghdos, line 182: it can be set to e.g., ``-D GPU_TARGETS="gfx1032;gfx1035"``

docs/understand/cmake_packages.rst
```diff
@@ -171,4 +220,156 @@ ROCm CMake Packages
...
+===================
+
+CMake command-lines depending on how specific users like to be when compiling
+code can grow to unwieldy lengths. This is the primary reason why projects tend
+to bake script snippets into their build definitions controlling compiler
+warning levels, changing CMake defaults (``CMAKE_BUILD_TYPE`` or
+``BUILD_SHARED_LIBS`` just to name a few) and all sorts anti-patterns, all in
+the name of convenience.
+
+Load on the command-line interface (CLI) starts immediately by selecting a
```
arghdos, line 235: This is unclear: I think we're saying that the `toolchain` file (and various variables, such as `CMAKE_CXX_COMPILER`) are set _before_ the beginning of the user's `CMake` file?

arghdos (changes requested):
This is a really good start!

Left some comments / clarification requests.

arghdos:
cc @ye-luo, would be interested in any of your feedback as a user as well.

docs/understand/cmake_packages.rst
saadrahim: Let's add this request to rocm-examples

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
-   ``CMAKE_C_COMPILER`` and ``CMAKE_CXX_COMPILER`` variable or using the ``CC``
-   and ``CXX`` environment variables. This can be set when configuring CMake or
-   put into a CMake toolchain file. The device compiler must be set to a
-   compiler that supports AMD GPU targets, which is usually Clang.
+Consuming the HIP API in C++ code
+---------------------------------
 
-The ``find_package(hip)`` provides the ``hip::device`` target to add all the
-flags for device compilation
+Use the HIP API without compiling the GPU device code. As there is no GPU code,
```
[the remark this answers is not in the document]
MathiasMagnus: AFAIK yes, but if anyone can confirm, please chime in. [Some comments](https://github.com/ROCm-Developer-Tools/hipamd/blob/4209792929ddf54ba9530813b7879cfdee42df14/include/hip/amd_detail/amd_hip_bfloat16.h#L41-L42) in code imply that any C compiler is enough, and even C++ compilers below C++11 are intended to work, though setting C++11 as a minimal standard is advised.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
--  Append the variable in the CMakeLists.txt file. This is useful if the
-   dependencies are found in a common location. For example, when the binaries
-   provided on `repo.radeon.com <http://repo.radeon.com>`_ are installed to
-   */opt/rocm*, you can add the following line to a CMakeLists.txt file
+::
 
-   ::
+    cmake_minimum_required(VERSION 3.21) # enable_language(HIP)
+    cmake_policy(VERSION 3.21...3.27)
+    project(MyProj LANGUAGES CXX HIP)
```
cgmb, line 76: The `CXX HIP` thing turned out to be a big miscommunication. This can just be `project(MyProj LANGUAGES HIP)`.

docs/understand/cmake_packages.rst
cgmb: It should also be `3.21.3`.

docs/understand/cmake_packages.rst
MathiasMagnus: Removed.

docs/understand/cmake_packages.rst
MathiasMagnus: Fixed.

docs/understand/cmake_packages.rst
MathiasMagnus: Fixed.

docs/understand/cmake_packages.rst
MathiasMagnus: Fixed.

docs/understand/cmake_packages.rst
MathiasMagnus: Once when example is made, we can extract the relevant parts here, or just link to the sample. My hope is that eventually we can retire that Module by adding its functionality to the HIP language support.

Just look at any of the HIP-Basic samples of ROCm-examples ([saxpy](https://github.com/amd/rocm-examples/blob/develop/HIP-Basic/saxpy/CMakeLists.txt) for eg.) or any of the HIP library samples that also have device code (hipCUB [device_sum](https://github.com/amd/rocm-examples/blob/develop/Libraries/hipCUB/device_sum/CMakeLists.txt) for eg.). They all properly support both HIP on AMD and HIP on NV and the build definition is a series of jumping through flaming hoops. I'm certain we could do better than that.

docs/understand/cmake_packages.rst
MathiasMagnus: rocPRIM, hipBLAS, etc. Should we clarify?

docs/understand/cmake_packages.rst
MathiasMagnus: Suggestion applied.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
 
-   # Search for rocm in common locations
-   list(APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
-   # Find hip
-   find_package(hip)
-   # Create the library
-   add_library(myLib ...)
-   # Link with HIP
-   target_link_libraries(myLib hip::host)
+    cmake_minimum_required(VERSION 3.5) # find_package(miopen)
```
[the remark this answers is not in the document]
MathiasMagnus: It's rationale on why a certain minimum version is requried. The introduction (and most libraries I checked) seem to be requiring 3.5 minimally.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
-  list(APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
-  # Find hip
-  find_package(hip)
-  # Create library
-  add_library(myLib ...)
-  # Link with HIP
-  target_link_libraries(myLib hip::device)
-
-This project can then be configured with
+    cmake_minimum_required(VERSION 3.5) # find_package(hip)
```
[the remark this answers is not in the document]
MathiasMagnus: Same here.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
+The ``find_package(hip)`` provides the ``hip::device`` imported target to add
+all the flags necessary for device compilation.
 
 ::
 
-    cmake -DCMAKE_C_COMPILER=/opt/rocm/llvm/bin/clang -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++ ..
-
-Which uses the device compiler provided from the binary packages from
-`repo.radeon.com <http://repo.radeon.com>`_.
+    cmake_minimum_required(VERSION 3.8) # cxx_std_11
```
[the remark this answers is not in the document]
MathiasMagnus: Would `# using cxx_std_11 below` be a clearer comment? (Applies to other cases as well.)

docs/understand/cmake_packages.rst
MathiasMagnus: Fixed.

docs/understand/cmake_packages.rst
MathiasMagnus: What this is saying, is that supplying a complete toolchain on the CLI requires setting 2-10 variables. To ease providing all these variables, CMake reads a few env vars, and also allows reading an entire script, the toolchain file. But the toolchain is just the tip of the iceberg, because there are many other variables the user may want to set.

docs/understand/cmake_packages.rst
MathiasMagnus: 3.21.3 added.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,188 @@
...
 
-    list (APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
+CMake itself only hosts part of the HIP language support, such as defining
+HIP-specific properties, etc. while the other half ships with the HIP
+implementation, such as ROCm. CMake will search for a file
+`hip-lang-config.cmake` describing how the the properties defined by CMake
+translate to toolchain invocations. If one installs CMake into non-standard
+locations or has multiple HIP toolchains installed side-by-side and wants to
+instruct CMake to choose a specific one, it can be done by setting
+``-D CMAKE_HIP_COMPILER_ROCM_ROOT:PATH=`` to the root of the ROCm installation.
```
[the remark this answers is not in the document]
ye-luo: @MathiasMagnus what is `hipconfig`?
My take, if `CMAKE_HIP_COMPILER_ROCM_ROOT` is intended to find `hip-lang.cmake`. The poor choice of name already keeps users away from it. Setting `CMAKE_CXX_COMPILER` and expect things to work is not specific to any CXX compiler. So a CMake user has the right to expect `CMAKE_HIP_COMPILER` not being a ROCm specific thing. ROCm should work with `CMAKE_HIP_COMPILER`. Whether other HIP implementation works or not is a thing between that implementation (not ROCm) and CMake.

docs/understand/cmake_packages.rst
```diff
@@ -1,139 +1,183 @@
...
-There are two ways to use HIP in CMake:
+::
 
--  Use the HIP API without compiling the GPU device code. As there is no GPU
-   code, any C or C++ compiler can be used. The ``find_package(hip)`` provides
-   the ``hip::host`` target to use HIP in this context
+    cmake_minimum_required(VERSION 3.21) # enable_language(HIP)
+    cmake_policy(VERSION 3.21.3...3.27)
+    project(MyProj LANGUAGES HIP)
+    add_executable(MyApp Main.hip)
```
ye-luo, line 76: Considering users may have pre-existing cuda source files. Adding an example can be helpful.
```
add_library(foo foo.cu)
set_source_files_properties(foo.cu PROPERTIES LANGUAGE HIP)
```

docs/understand/cmake_packages.rst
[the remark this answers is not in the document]
arghdos: Yes -- either:

```
# using cxx_std_11 below
```

or

```
 v3.8 required for cxx_std_11
```

docs/understand/cmake_packages.rst
arghdos: Yeah, adding an "e.g., rocPRIM or other ROCm libraries" would clear it up.

MathiasMagnus: Clarified.

docs/understand/cmake_packages.rst
MathiasMagnus: Added.

docs/understand/cmake_packages.rst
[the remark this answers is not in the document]
MathiasMagnus: Fixing globally

saadrahim:
This PR will be merge soon unless further concerns are raised. All outstanding comments have been addressed.

docs/understand/cmake_packages.rst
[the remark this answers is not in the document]
arghdos: I believe that for linking against the HIP runtime, you should be able to use plain C, or C++11 or a mix

arghdos:
LGTM

ye-luo:
I still prefer recommending `CMAKE_HIP_COMPILER` to `CMAKE_HIP_COMPILER_ROCM_ROOT`. All the rest looks good to me.

saadrahim (approved):

PR #2157, opened 2023-05-22
Update openmp.md

SwRaw:
Modified the link for GPU support from docs.amd.com to RTD.


Rmalavally (approved):
Looks good, Swati. Nice :)

saadrahim (changes requested):
Please use relative links to the file associated. The full URL is not required and is harder to maintain.

PR #2229, opened 2023-06-08
Install extra repos

Maetveis:
Add instructions for adding extra repositories in RHEL and SLES
    
The hip-devel package depends on perl modules not distributed by default
on RHEL and SLES distriubutions, these can be installed from EPEL and
the `devel:languages:perl` repository respectively.

Ideally in the future these dependencies would be replaced with packages
available from default repositories, but in the meanwhile this should
be at least documented.

Based on #2228 to avoid conflicts, only the last commit is new in this PR


Maetveis:
Ci should get fixed by: https://github.com/RadeonOpenCompute/rocm-docs-core/pull/236

saadrahim (approved):

PR #2236, opened 2023-06-10
Fixing HIP link

saadrahim:


Rmalavally (commented):
HIP API Reference Manual

samjwu (approved):

saadrahim:
> HIP API Reference Manual

It is part of the grid card item identifying it as HIP. There is no need to repeat HIP.

PR #2237, opened 2023-06-11
Update tensorflow_install.md

alfinauzikri:
Fixed writing commands so that when executed by copy paste it doesn't cause an error.

### DIFF
--- a/docs/how_to/tensorflow_install/tensorflow_install.md
+++ b/docs/how_to/tensorflow_install/tensorflow_install.md
@@ -33,9 +33,7 @@ Follow these steps:
 2. Once you have pulled the image, run it by using the command below:
 
    ```bash
-   docker run -it --network=host --device=/dev/kfd --device=/dev/dri
-   --ipc=host --shm-size 16G --group-add video --cap-add=SYS_PTRACE
-   --security-opt seccomp=unconfined rocm/tensorflow:latest
+   docker run -it --network=host --device=/dev/kfd --device=/dev/dri --ipc=host --shm-size 16G --group-add video --cap-add=SYS_PTRACE --security-opt seccomp=unconfined rocm/tensorflow:latest
    ```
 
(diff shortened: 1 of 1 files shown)


docs/how_to/tensorflow_install/tensorflow_install.md
```diff
@@ -33,9 +33,7 @@ Follow these steps:
 2. Once you have pulled the image, run it by using the command below:
 
    ```bash
-   docker run -it --network=host --device=/dev/kfd --device=/dev/dri
-   --ipc=host --shm-size 16G --group-add video --cap-add=SYS_PTRACE
-   --security-opt seccomp=unconfined rocm/tensorflow:latest
+   docker run -it --network=host --device=/dev/kfd --device=/dev/dri --ipc=host --shm-size 16G --group-add video --cap-add=SYS_PTRACE --security-opt seccomp=unconfined rocm/tensorflow:latest
```
saadrahim, line 36: Instead of combining into a single line, it would be better to keep the original 3 separate lines and use the \ to signify a line break in bash. If you wish to correct it, please do so or I can make an alternate PR.

saadrahim (changes requested):

docs/how_to/tensorflow_install/tensorflow_install.md
saadrahim: Thank you for pointing out this issue.

Rmalavally: @amathews-amd FYI. Please review.

alfinauzikri: Ok, I've followed your advice @saadrahim 

Thank You..

alfinauzikri (commented):
Following @saadrahim's suggestion of using backslash (\\) to signify a line break in bash.

PR #2245, opened 2023-06-15
Swati develop

SwRaw:


docs/reference/openmp/openmp.md
samjwu, line 1: The spell checker doesn't recognize some words that were added. They will have to be added to the `.wordlist.txt` file.

Example:
```
# openmp
ICV
Multithreaded
```

SwRaw: I think the word added to .wordlist.txt is multi-threaded. I believe it should be multithreaded instead. If we all agree then this should be changed else I will have to change the spelling in my document.

samjwu: Sounds good

saadrahim (approved):

PR #2267, opened 2023-06-23
Announcements Template

saadrahim:
Trying out the look for making announcements in a manner separate to the release notes where we can go deeper into a topic.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 # AMD ROCm™ Documentation
 
-:::::{grid} 1 1 3 3
+:::::{grid} 1 1 4 4
 :gutter: 1
 
 ::::{grid-item}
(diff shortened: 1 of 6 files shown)


docs/index.md
```diff
@@ -34,6 +34,15 @@ agile, flexible, rapid and secure manner. [more...](rocm)
 :::
 ::::
 
+::::{grid-item}
+:::{dropdown} [Announcements](announcements)
+
+- [Planned Deprecations](deprecations)
+- [Docs-As-Code](docs-as-code)
```
samjwu, line 41: is the plan to have separate pages for announcements and docs-as-code?

saadrahim: These were just sample announcements, could be anything here, maybe limit 5 and rest as an archive.

saadrahim:
#2428 will implement this

PR #2305, opened 2023-06-29
Add configurations for PDF output on Read the Docs

samjwu:
Sample on 5.5.1: https://rocm.docs.amd.com/_/downloads/en/docs-5.5.1/pdf/

Ref: https://www.sphinx-doc.org/en/master/usage/configuration.html

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,6 +14,10 @@ shutil.copy2('../RELEASE.md','./release.md')
 # Keep capitalization due to similar linking on GitHub's markdown preview.
 shutil.copy2('../CHANGELOG.md','./CHANGELOG.md')
 
+# for PDF output on Read the Docs
+project = "ROCm Documentation"
+author = "AMD"
+
 setting_all_article_info = True
 all_article_info_os = ["linux"]
 all_article_info_author = ""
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -14,12 +14,22 @@
 # Keep capitalization due to similar linking on GitHub's markdown preview.
 shutil.copy2('../CHANGELOG.md','./CHANGELOG.md')
 
+# for PDF output on Read the Docs
+project = "ROCm Documentation"
+author = "AMD"
```
saadrahim, line 19: The author is AMD Inc. not just AMD

samjwu: Done in [e82fcbb](https://github.com/RadeonOpenCompute/ROCm/pull/2305/commits/e82fcbb107ad8254910d57f82abbb06613f5812d)

Do we also want to add a copyright?

saadrahim: You should add that as well.

samjwu: [add copyright to pdf](https://github.com/RadeonOpenCompute/ROCm/pull/2305/commits/a3ec07e634775023bb537b307db72b2b8ee6e5a1)

saadrahim (approved):

PR #2338, opened 2023-07-27
Create using_gpu_sanitizer.md

Rmalavally:
Draft documentation for using GPU sanitizer

PR #2348, opened 2023-07-27
Update .wordlist.txt

Rmalavally:
Updated word list

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -103,3 +103,9 @@ DGEMM
 HPCG
 HPL
 IOPM
+amdclang
+asan
+complilation
+ldd
+libclang
+readelf


.wordlist.txt
```diff
@@ -103,3 +103,9 @@ DGEMM
 HPCG
 HPL
 IOPM
+amdclang
+asan
+complilation
```
saadrahim, line 108: compilation, this line shouldn't be added

Rmalavally: Agreed. Please remove 'compilation'.

saadrahim: You need to remove the line.

samjwu (changes requested):
Now that https://github.com/RadeonOpenCompute/ROCm/pull/2338 is passing, I don't think we should add these to the word list.

We want these keywords to be wrapped in tilde.

.wordlist.txt
samjwu, line 1: These words were being flagged intermittently:

uninstrumented
detections

samjwu:
Fixed/added in https://github.com/RadeonOpenCompute/ROCm/pull/2338

PR #2361, opened 2023-08-02
Test add myst-parser html metadata

samjwu:
global metadata (all files) - conf.py

local metadata (per file) - markdown file (index.md)

re: https://myst-parser.readthedocs.io/en/latest/configuration.html#setting-html-metadata

image


saadrahim:
I think this should wait until we are ready with proper meta classification system. As a POC, we have reached our goals.

Issue #2399, opened 2023-08-23
Incorrect ROCm 5.5 install instructions


cgmb:
Following the [ROCm 5.5 install instructions](https://rocm.docs.amd.com/en/docs-5.5.1/deploy/linux/installer/install.html) will result in installing ROCm 5.6.

For example:

```
sudo apt update
wget https://repo.radeon.com/amdgpu-install/5.6/ubuntu/focal/amdgpu-install_5.6.50600-1_all.deb
sudo apt install ./amdgpu-install_5.6.50600-1_all.deb
```

Promesis:
So it should be this?

```shell
sudo apt update
wget https://repo.radeon.com/amdgpu-install/5.5.1/ubuntu/focal/amdgpu-install_5.5.50501-1_all.deb
sudo apt install ./amdgpu-install_5.5.50501-1_all.deb
```

Promesis:
A lot of mistakes. RHEL and SUSE.

Promesis:
I incorrectly edited it on `develop` branch. It should be `docs/5.5.1`.

cgmb:
Thanks @Promesis!

PR #2403, opened 2023-08-24
	modified:   install.md

Promesis:
Fixed issue #2399.

### DIFF
--- a/docs/deploy/linux/installer/install.md
+++ b/docs/deploy/linux/installer/install.md
@@ -18,7 +18,7 @@ following commands based on your distribution.
 
 ```shell
 sudo apt update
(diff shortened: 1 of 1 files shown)


cgmb:
@saadrahim, it would be nice to set up a CI system for the install instructions. In rocSOLVER, we test our examples compile by using [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude), so the actual code can be in a file that is part of our normal build. e.g.,

```rst
.. literalinclude:: ../../clients/samples/example_basic.c
    :language: c
```

We could put the install instructions into some actual bash scripts so that the CI can spin up a fresh docker container and run them. Then just check that ROCm is installed after the scripts have been run.

PR #2409, opened 2023-08-27
Create CODE_OF_CONDUCT.md

saadrahim:


samjwu (approved):

dsclear-amd:
As a suggestion: the proposed code of conduct only calls for permanent bans in case of a _pattern_ of bad behaviour, but this suggests that even the most egregious behaviour can get a temporary ban at best (which is probably not desired).

PR #2426, opened 2023-08-31
Fix RHEL9 installer filename links

prckent:
The el8 references were not changed to el9. I verified the updated filenames exist.


saadrahim:
These fixes may have been fixed in the rocm-5.6 branches. Let me merge the differences and see if this is still needed.

saadrahim (approved):
Thank you for the contribution.

PR #2441, opened 2023-09-08
Table of Contents Updates

LisaDelaney:


docs/about/release_notes.md
saadrahim, line 1: File is probably does not need to be added. I think you can add to the .gitnore list. You should move it in conf.py.

Rmalavally:
@LisaDelaney Good effort, Lisa. Quick question. Do we want 'ai' as a prefix or a suffix? We have 'ai.libraries' and 'rocm.ai'. Is this intended?

saadrahim (approved):

PR #2447, opened 2023-09-11
Main page updates

LisaDelaney:


saadrahim:
What about buttons for the section below? https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html

We are not really using the dead space in the right
image

PR #2459, opened 2023-09-15
Update compatibility matrix

Naraenda:


Rmalavally:
@Naraenda thanks for the updates.  How are we indicating the end of support for MI50 (collectively gfx906 GPUs) in the 5.7 release?

PR #2484, opened 2023-09-19
Specify font package for xelatex engine for pdf

samjwu:
https://www.sphinx-doc.org/en/master/latex.html#module-latex

issue: xelatex uses GNU FreeFont, which was not available


samjwu:
Covered by https://github.com/RadeonOpenCompute/ROCm/pull/2485/

PR #2485, opened 2023-09-19
Fix PDF doc builds

samjwu:
RE: Error: Failed to load PDF document
image

Specify font package for xelatex engine for pdf
Ref: https://www.sphinx-doc.org/en/master/latex.html#module-latex

Specify image size for svg


Rmalavally (dismissed):

saadrahim (changes requested):
I think there are formatting and potentially linting issues with this. Please check again.

docs/conf.py
```diff
@@ -14,7 +14,7 @@
 # Keep capitalization due to similar linking on GitHub's markdown preview.
 shutil.copy2('../CHANGELOG.md','./CHANGELOG.md')
 
-latex_engine = "xel  atex"
+latex_engine = "xelatex"
```
Rmalavally, line 17: @samjwu We should add this word to the wordlist

samjwu:
redone in https://github.com/RadeonOpenCompute/ROCm/pull/2486

docs/conf.py
samjwu: this is in conf.py, which is not checked

Rmalavally: Sorry, this comment wasn't for this PR. Will delete. 

Thanks for clarifying, Sam.

PR #2509, opened 2023-09-27
Edited the quickstart guide for linux so all commands appear in a single place

abhimeda:
Edited the file and moved all commands to a singular console block that can be copy pasted into a terminal to install ROCm.

### DIFF
--- a/docs/tutorials/quick-start/linux.md
+++ b/docs/tutorials/quick-start/linux.md
@@ -36,7 +36,10 @@ $ sudo reboot
 :sync: ubuntu-22.04-IS
 
 ```console
-$ temp
+$ sudo apt update 
+$ wget https://repo.radeon.com/amdgpu-install/5.7/ubuntu/jammy/amdgpu-install_5.7.50700-1_all.deb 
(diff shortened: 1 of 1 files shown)


samjwu (commented):
The explanations may be helpful for users installing (eg: the {rubric} section)

I think we should keep those as comments or something similar

Rmalavally:
> The explanations may be helpful for users installing (eg: the {rubric} section)
> 
> I think we should keep those as comments or something similar

Thanks, Sam. 

My recommendation is to have QA review it before we merge the PR. @rkebichi

rkebichi:
> > The explanations may be helpful for users installing (eg: the {rubric} section)
> > I think we should keep those as comments or something similar
> 
> Thanks, Sam.
> 
> My recommendation is to have QA review it before we merge the PR. @rkebichi

Hi @Rmalavally, I checked with Freddy from installation and packaging team and he said "testing it is good to ensure formatting do not cause issues". Please do the necessary to have this doc change tested. Thanks.

abhimeda:
Could I receive an update with the testing? Is there anything I need to clarify or explain?

PR #2532, opened 2023-10-06
Move spack & update pytorch

LisaDelaney:


docs/install/pytorch-install.md
```diff
@@ -136,11 +137,11 @@ table, choose ROCm from the _Compute Platform_ row.
    After installing ROCm PyTorch wheels, run the following code:
 
    ```bash
-   #Optional; replace 'gfx90a' with your architecture
+   #Optional; replace 'gfx90a' with your architecture and 5.6 with your preferred ROCm version
```
Rmalavally, line 140: @LisaDelaney Did we mean to mention the version number here?

Rmalavally: Thanks, Lisa. Please check if the example can be generalized or if it is tied to a specific release. Is the example compatible with other ROCm versions? Is there a requirement?

LisaDelaney: I think it's probably more helpful to have a version number in the example (the example starts on line 139 if you want to review it), rather than, say "export ROCM_VERSION=x.x". We also mention specific architectures. These steps could, in theory, change with future versions (these instructions list versions `5.4`, `5.5`, and`5.6` specifically...so I'm guessing they may not apply to every version).

Rmalavally: Sounds good. We want to confirm that the example works with the 'preferred ROCm version'. If you have the confirmation from the SME, we are good.

saadrahim: Maybe you can leave 5.6 in the export ROCM_VERSION=5.6 and tell the user to change the ROCM_VERSION variable to your desired ROCm version. That way we don't have to maintain this page all the time. However, this may be a case to use Jinja in the future to both maintain the latest version number and test the code snippet.

saadrahim (approved):

PR #2546, opened 2023-10-12
Using GPU Sanitizer 

Rmalavally:
Using GPU Sanitizer for ROCm 5.7.1


samjwu (changes requested):
looks good, just need to resolve merge conflicts

.wordlist.txt
LisaDelaney, line 1: This commit is trying to 'revert' changes we recently made to this file. It's appending the old version of the wordlist to the new version.

PR #2585, opened 2023-10-19
Structure cleanup

LisaDelaney:


docs/CHANGELOG.md
samjwu, line 1: This file is overwritten by CHANGELOG.md when the docs build

Not sure why we are still tracking this. It can be safely removed

RE: https://github.com/RadeonOpenCompute/ROCm/blob/roc-5.7.x/docs/conf.py#L33C14-L33C14

saadrahim (approved):

PR #2660, opened 2023-11-20
update ROCm name

LisaDelaney:


CHANGELOG.md
```diff
@@ -11,7 +11,7 @@
 
 <!-- spellcheck-disable -->
 
-This page contains the release notes for the ROCm platform.
+This page contains the release notes for AMD ROCm Software.
```
saadrahim, line 14: software should be lowercased except in titles (I know not really sentence case).

saadrahim (changes requested):

CHANGELOG.md
Rmalavally: Thanks, Lisa and Saad. We used to have AMD ROCm before. However, we switched to ROCm (removed AMD) during a rebranding exercise. I prefer ROCm, but I am open to all options. I suggest checking with Bryce.

saadrahim: > AMD ROCm Software [formal reference] versus ROCm software [casual reference]
This makes sense and we can stick with it.

saadrahim (approved):

PR #2734, opened 2023-12-15
Delete docs/about/compatibility/linux-support.md

saadrahim:


samjwu (changes requested):
will also want to remove the entry in the toc

samjwu (approved):

PR #2738, opened 2023-12-16
Corrected OS version

pramenku:
There is no 22.04.5 exist.
It's 22.04.3 which has been tested and supported


Rmalavally (approved):
Thanks so much.

saadrahim (approved):

PR #2772, opened 2024-01-02
Abhimeda adding workflow file to create GitHub Action

abhimeda:
I am tasked with adding a GitHub action to process incoming GitHub issues. The AMD GitHub admin team asked me to try out one of their runners and to do this I need to load in a workflow file.


saadrahim (dismissed):
Recommend moving the workflow repository to the ROCm organization.

saadrahim (approved):

PR #2795, opened 2024-01-10
Revert "Update CONTRIBUTING.md"

LisaDelaney:
Reverts ROCm/ROCm#2791 --> the link needs to remain as it was because we have a script that copies this file into the docs folder and it is this file that is used to build the docs. If you go to the ROCm docs site, you'll see the link was working (and this PR broke it).


saadrahim (approved):
we need some processing when building the website to fix this correctly.

samjwu (approved):

PR #2798, opened 2024-01-11
Update Contributing.md to direct PRs to use repo's default branch

yhuiYH:
Small text change to have PRs target default branch, not just develop branch.  To cover those repos that don't use develop.


Rmalavally (approved):
Makes sense, and looks good. 

Thanks so much.

samjwu (approved):

PR #2834, opened 2024-01-23
Add codeowners for documentation

samjwu:
Adding documentation team as codeowners for documentation.
Goal is to make this group of code owners reviewers when merging documentation changes. The header directory is related to documentation and thus changes to that are also flagged for documentation review.

**Please note that this requires the team to be added as a role with Write permissions.**


Rmalavally (approved):

randyh62 (approved):
Looks good to me.

PR #2887, opened 2024-02-08
update tm

LisaDelaney:


saadrahim (approved):

samjwu (commented):
Is the trademark only needed the first time it appears in a doc?

Rmalavally:
> Is the trademark only needed the first time it appears in a doc?

First instance on every 'page'.

samjwu (approved):

PR #2910, opened 2024-02-20
add alt text

LisaDelaney:


saadrahim:
I don't see the alt text when I hover the mouse over the images. Do you see them?

LisaDelaney:
> I don't see the alt text when I hover the mouse over the images. Do you see them?

I don't...not sure if that impacts what a screen reader picks up...

samjwu:
```
<img alt="Install documentation" class="sd-card-img-top" src="_images/banner-installation.jpg">
```

The HTML looks correct. The alt text is only displayed when the image cannot be loaded, not on hover.

samjwu (approved):

PR #2929, opened 2024-02-26
Define Submodules from the Manifest file.

dgaliffiAMD:
Synced to ROCm 6.0.2


dgaliffiAMD:
Convert to "draft". Just testing.

saadrahim:
According to pitchfork, a slightly different structure maybe more intuitive, instead of everything in the root folder.

dgaliffiAMD:
> According to pitchfork, a slightly different structure maybe more intuitive, instead of everything in the root folder.

I'll move them to `libs/`

saadrahim (approved):

PR #2966, opened 2024-03-20
redirect openmp links to cover the move

parbenc:
https://pypi.org/project/sphinx-reredirects/

https://documatt.com/sphinx-reredirects/usage.html


samjwu (approved):
confirmed https://advanced-micro-devices-demo--2966.com.readthedocs.build/en/2966/reference/openmp/openmp.html redirects correctly

yhuiYH (approved):

yhuiYH:
Hmm... thx for that detail @MKKnorr  perhaps we should re-evaluate.  I did ping @samjwu on this, and he did mention it was the first instance of using this extension.

MKKnorr:
> @MKKnorr @yhuiYH as far as I know the readthedocs redirect is limited based on the license, so I thought best to go with the sphinx version

Fair point, that problem already popped up once, but was due to automatically generated redirects from a script (i.e. several pages from the same directory were individually redirected to the new page, instead of redirecting the whole directory). Currently we still have about 300 redirects left.

On top of that, this would open up another discussion, that doesn't belong here: how should we handle moving directories/pages in general. It is not feasible to keep **all** redirects from old pages, especially as the directory structure currently is not very stable, and probably subject to even more change (e.g. why is the `compatibility` directory now in the `about` directory? This doesn't make any sense, and the `about` directory is not even exposed anywhere, neither on the landing page nor in the toc)

PR #2967, opened 2024-03-20
AMD SMI changelog

samjwu:


Rmalavally (commented):
@samjwu This changelog is in a different format from others. Do we want to pull this in?

samjwu:
Covered by Roopa

PR #2991, opened 2024-04-02
Update using-gpu-sanitizer.md

Rmalavally:
Minor OpenMP update


yhuiYH (approved):
OK, looks fine.  I think giving an example of `rocm-<version>` is better, like calling out `rocm-6.0.2`, but hopefully the reader can figure that out.

Rmalavally:
> OK, looks fine. I think giving an example of `rocm-<version>` is better, like calling out `rocm-6.0.2`, but hopefully the reader can figure that out.

I agree. However, from a documentation perspective, specifying a version unless absolutely required is not recommended due to maintenance concerns.

PR #2992, opened 2024-04-02
Update links

LisaDelaney:


LisaDelaney:
the markdown linting can be ignored...it doesn't like that I used reference tags within the card grid, but it renders fine.

samjwu:
In general this rule should be followed to avoid broken links, but in this case we can disable MD051 since the grid item card does not have an anchor to point to without using Sphinx Markdown anchor point syntax

docs/about/compatibility/precision-support.rst
```diff
@@ -1,13 +1,14 @@
...
   :description: Supported data types in ROCm
-  :keywords: int8, float8, float8 (E4M3), float8 (E5M2), bfloat8, float16, half, bfloat16, tensorfloat32, float, float32, float64, double, AMD, ROCm, AMDGPU
-
-.. _rocm-supported-data-types:
+  :keywords: int8, float8, float8 (E4M3), float8 (E5M2), bfloat8, float16, half, bfloat16, tensorfloat32, float,
+   float32, float64, double, AMD, ROCm, AMDGPU
 
 *************************************************************
-ROCm data type specifications
+Precision support
```
Rmalavally, line 7: @yhuiYH Should we consider renaming the heading to 

ROCm-supported data and HIP types

yhuiYH: No, I think it's fine if we keep it as "Precision support"

lpaoletti:
Let's remove the hyphen here.
Also remove the (Linux) as this is in the Linux box.


Linux

      Quick-start (Linux)

Rmalavally:
> Let's remove the hyphen here. Also remove the (Linux) as this is in the Linux box.
> 
> Linux
> 
> ```
>   Quick-start (Linux)
> ```

let me look into this today, Leo.

docs/about/compatibility/precision-support.rst
```diff
@@ -1,13 +1,14 @@
...
 *************************************************************
 
+Use the following sections to identify which data types and HIP types are supported with ROCm.
```
lpaoletti, line 10: Use the following sections to identify data types and HIP types ROCm supports.

lpaoletti (approved):

docs/about/compatibility/precision-support.rst
yhuiYH: fixed

yhuiYH (approved):

lpaoletti (approved):

PR #2994, opened 2024-04-02
refactor(.markdownlint-cli2.yaml): Removed old and unused mdlint config

samjwu:
replaced by the one in rocm-docs-core (https://github.com/ROCm/rocm-docs-core/blob/develop/.markdownlint.yaml)


samjwu:
https://github.com/DavidAnson/markdownlint-cli2

There are two kinds of configuration file (both detailed below):
Configuration files like .markdownlint-cli2.* allow complete control of markdownlint-cli2 behavior and are also used by vscode-markdownlint.
If multiple of these files are present in the same directory, only one is used according to the following precedence:
.markdownlint-cli2.jsonc
.markdownlint-cli2.yaml
.markdownlint-cli2.cjs
.markdownlint-cli2.mjs
package.json (only supported in the current directory)
Configuration files like .markdownlint.* allow control over only the markdownlint config object and tend to be supported more broadly (such as by markdownlint-cli).
If multiple of these files are present in the same directory, only one is used according to the following precedence:
.markdownlint.jsonc
.markdownlint.json
.markdownlint.yaml
.markdownlint.yml
.markdownlint.cjs
.markdownlint.mjs

PR #3018, opened 2024-04-16
Add additional autotag script note to avoid duplicating data in changelog template

samjwu:
IMPORTANT: It is key to update the template Markdown files in `tools/autotag/templates/rocm_changes` (eg: `5.6.0.md`) and not the `CHANGELOG.md` itself to ensure that updates are not overwritten by the autotag script. The template should only have content from changelogs that are not included by the script to avoid duplicating data.

Refer to existing templates for examples (eg: 6.0.0.md)

### DIFF
--- a/tools/autotag/README.md
+++ b/tools/autotag/README.md
@@ -13,7 +13,7 @@
 
 ## Updating the changelog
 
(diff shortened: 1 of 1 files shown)


yhuiYH (approved):
Thx for the updated instructions Sam.

PR #3082, opened 2024-05-04
Add compatibility matrix

yhuiYH:

PR #3086, opened 2024-05-06
Add compatibility matrix (#3082)

yhuiYH:
* add compatibility matrix and custom css

* fix toc

* reorder some components in matrix, add missing tools to reference page

* Update docs/compatibility/compatibility-matrix.rst

---------

### DIFF
--- a/docs/compatibility/compatibility-matrix.rst
+++ b/docs/compatibility/compatibility-matrix.rst
@@ -15,11 +15,11 @@ Use this matrix to view the ROCm compatibility across successive major and minor
       :header: "ROCm Version", "6.1.0", "6.0.0"
       :stub-columns: 1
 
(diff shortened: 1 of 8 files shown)


samjwu (approved):
looks good

the notation may be confusing to some users

Ubuntu 22.04.z (z=3-4)
vs
Ubuntu 22.04.3
Ubuntu 22.04.4

yhuiYH:
> looks good
> 
> the notation may be confusing to some users
> 
> Ubuntu 22.04.z (z=3-4) vs Ubuntu 22.04.3 Ubuntu 22.04.4

I agree with you and also voiced my concern, but I was asked to communicate it this way.  I prefer 22.04.3 better.  It's more clear, and searchable.

PR #3115, opened 2024-05-10
Add build pools, change gpu targets, and build tests

amd-jmacaran:
To best determine hardware specs per repo, added more build pool options with varying number of vCPUs, RAM size, etc. and will kick off builds with test targets enabled to determine long-term cost values.

pipelines_repo reference not necessary for tag-builds. Removing the references makes testing/debugging easier.


alexxu-amd (approved):

amd-jmacaran:
After refactors, all components building with test binaries.

.azuredevops/tag-builds/AMDMIGraphX.yml
```diff
@@ -1,6 +1,6 @@
 variables:
 - group: common
-- template: /.azuredevops/variables-global.yml@pipelines_repo
```
alexxu-amd, line 3: We removed all pipelines_repo usage but still define it. Is it refactored somewhere?

.azuredevops/components/rocminfo.yml
```diff
@@ -5,14 +5,18 @@ parameters:
...
+    - ROCR-Runtime
 
 jobs:
 - job: rocminfo
   variables:
   - group: common
   - template: /.azuredevops/variables-global.yml
   pool:
-    vmImage: ${{ variables.LOW_END_BUILD_POOL }}
+    vmImage: ${{ variables.BASE_BUILD_POOL }}
```
alexxu-amd, line 19: I think BASE_BUILD_POOL is a local machine. Wonder if that's desired

amd-jmacaran: BASE_BUILD_POOL is not a local machine. It is the ubuntu-22.04, Microsoft-hosted pool. This is desired. ON_PREM_BUILD_POOL are local machines.

.azuredevops/tag-builds/AMDMIGraphX.yml
amd-jmacaran: Where do you see it defined? It is implied that the current repo is the pipelines_repo (i.e., ROCm/ROCm).

alexxu-amd: I'm a little confused about this, pipelines_repo seems like a variable defined by us. The first time it appears is in tag-builds/*.yml under resources section (please correct me if I'm wrong). So I thought this was the definition and later when we do @pipelines_repo, we specifying repo as ROCm/ROCm.

alexxu-amd (approved):

PR #3137, opened 2024-05-16
Add pipelines for third party components

alexxu-amd:
Add the following third party components:

- lapack
- gtest
- grpc

### DIFF
--- a/.azuredevops/tag-builds/lapack.yml
+++ b/.azuredevops/tag-builds/lapack.yml
@@ -11,7 +11,6 @@ resources:
 
 trigger: none
 pr: none
-ci: none
 
 jobs:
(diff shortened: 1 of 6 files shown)


.azuredevops/tag-builds/lapack.yml
```diff
@@ -2,28 +2,15 @@ variables:
 - group: common
 - template: /.azuredevops/variables-global.yml
```
amd-jmacaran, line 4: You could add the lapackVersion parameter here and feed it as a parameter to the call on line 16. Then, you're able to specify the lapackVersion when manually kicking off a build in Azure Pipelines without having to make any code changes.

amd-jmacaran (approved):

PR #3140, opened 2024-05-20
Remove docker image dependency from some External CI

amd-jmacaran:
- Fixed compilers to pick amdclang.
- Added ldconfig step for setting up linking of shared libraries.
- Set Azure VMs to medium only.
- Remove empty directories in published tarballs.

### DIFF
--- a/.azuredevops/components/roctracer.yml
+++ b/.azuredevops/components/roctracer.yml
@@ -10,6 +10,8 @@ parameters:
   default:
     - cmake
     - ninja-build
+    - doxygen
+    - graphviz
 - name: pipModules
   type: object
   default:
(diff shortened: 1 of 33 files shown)


.azuredevops/components/roctracer.yml
```diff
@@ -56,9 +62,15 @@ jobs:
...
+        -DROCM_PATH=$(Agent.BuildDirectory)/rocm
+        -DCMAKE_MODULE_PATH=$(Agent.BuildDirectory)/rocm/lib/cmake/hip
+        -DCMAKE_PREFIX_PATH=$(Agent.BuildDirectory)/rocm
         -DGPU_TARGETS=gfx1030;gfx1100
         -GNinja
-  - template: ${{ variables.CI_TEMPLATE_PATH }}/steps/artifact-upload.yml
+#  - template: ${{ variables.CI_TEMPLATE_PATH }}/steps/artifact-upload.yml
+#  - task: Bash@3
+#    displayName: 'Tests'
+#    inputs:
```
alexxu-amd, line 73: Shall we delete comments for the dev branch?

amd-jmacaran: It's okay to leave it there for reference for when we add tests. The commands work but need to be run on a system with gpu.

PR #3151, opened 2024-05-23
External CI pipelines for openmp-extras repos

amd-jmacaran:
Separate tag-build pipelines for aomp and aomp-extras to unblock other work.


abhimeda (approved):
looks good

PR #3175, opened 2024-05-29
Update default.xml

perovskikh:


samjwu:
done in https://github.com/ROCm/ROCm/pull/3169

PR #3189, opened 2024-05-30
Fix a rocPRIM Build Flag from External CI

alexxu-amd:
Correct the path for -DCMAKE_CXX_COMPILER


alexxu-amd:
This change has already been included in another PR

PR #3197, opened 2024-05-30
Disable pdf builds

samjwu:
These are causing RTD builds to time out


randyh62 (approved):

randyh62:
Why no epub for the develop branch?

samjwu:
> Why no epub for the develop branch?

PDF and EPUB builds are significantly increasing build times (10-20 min each), which will cause the build to fail since the limit is 30 min

PR #3211, opened 2024-05-31
External CI: Remove redundant rocm_smi_lib pipeline ID

danielsu-amd:
There are two pipeline IDs for rocm_smi_lib, removed the one with hyphens since the component name does indeed have underscores.


amd-jmacaran (approved):
Good catch. LGTM

PR #3239, opened 2024-06-04
Fix first link in compatibility matrix table

samjwu:
Makes the [Operating Systems](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html) link appear in the table

Directly adds the header as text, to replace the :header: directive

### DIFF
--- a/docs/sphinx/static/css/rocm_custom.css
+++ b/docs/sphinx/static/css/rocm_custom.css
@@ -1,53 +1,6 @@
 
 /* Adds container for big tables, used for Compatibility Matrix */
 
-
-/* Header row to have opaque background colour when sticky */
-.format-big-table th {
-    background-color: var(--pst-color-background);
+.format-big-table {
     white-space: nowrap;
-  }
-  
-  /* Turn on borders for whole table */
-  .format-big-table th,
-  .format-big-table td {
-    border-width: 1px;
-  }
-  
-  /* .format-big-table th.head { */
-  /*   background-color: var(--pst-color-on-surface); */
-  /* } */
-  
-  /* Sticky header for table excluding the stub*/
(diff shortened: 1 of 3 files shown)


peterjunpark:
I think we should remove `rocm_custom.css` instead of this.

The custom css is incompatible with pydata-sphinx-theme's table update because they add `overflow-x:auto` to the table by default now. This conflicts with the `position:sticky; top:2rem;` header. This is why the header is covering the first row in the docs/6.1.2 build

docs/sphinx/static/css/rocm_custom.css
```diff
@@ -1,53 +1,6 @@
...
-  }*/
-  
-  /* Removes borders for stub column */
-  .format-big-table tbody th {
-    border-top: none;
-    border-bottom: none;
-  }
-  
-  /* For horizontal scrolling only.  Can't be combined with format-big-table container */
-  .horizontal-scrolling-container {
```
yhuiYH, line 51: But we still use .horizontal-scrolling-container for the large table.  I think we should keep that in the CSS.
Used here: https://raw.githubusercontent.com/ROCm/ROCm/develop/docs/reference/gpu-arch-specs.rst

peterjunpark: Oh I see. This custom css rule is actually redundant now -- `overflow-x: auto;` is added to tables by default after the update to `pydata-sphinx-theme`.

See https://pydata-sphinx-theme.readthedocs.io/en/v0.15.2/examples/kitchen-sink/tables.html#giant-tables vs https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/tables.html#giant-tables.

peterjunpark (approved):

PR #3274, opened 2024-06-10
Fix hipTensor build error on develop branch

alexxu-amd:
On the latest code, -GNinja breaks the build. Replace -GNinja with -j32 for multithreading


.azuredevops/templates/steps/build-cmake.yml
```diff
@@ -32,7 +35,7 @@ steps:
   displayName: '${{parameters.componentName }} Build'
   inputs:
     workingDirectory: ${{ parameters.cmakeBuildDir }}
-    cmakeArgs: '--build ${{ parameters.cmakeTargetDir }}'
+    cmakeArgs: '--build ${{ parameters.cmakeTargetDir }} -- ${{ parameters.multithreadFlag }}'
```
amd-jmacaran, line 38: What happens to other builds if multithreadFlag is empty, so the last part of the string is the double dash? Can you test another component with this branch, so it uses this template YAML file but everything else should be the same?

alexxu-amd: That's a good point. In this case, we should have the double dashes in the parameter

alexxu-amd: I moved the double dash into the parameter. Now other components will behave as before if multithreadFlag is empty

amd-jmacaran (approved):

PR #3275, opened 2024-06-10
Change composable_kernel CI pipeline to build for all architectures

alexxu-amd:
Add cmake flag to build instance only, since tests for different architectures cannot be built together.
Change pool to ultra and increase time limit to 3.5 hours.


amd-jmacaran (approved):
This looks good, but please do not merge until you submit and merge a PR on composable_kernel repo to disable PR builds until we figure out resourcing.

PR #3291, opened 2024-06-12
cherry-pick PR #3284

randyh62:
update ASan content


randyh62:
I had to update the contents of the using-gpu-sanitizer.md file more than just the cherry-pick was providing.

yhuiYH:
> I had to update the contents of the using-gpu-sanitizer.md file more than just the cherry-pick was providing.

That usually means there are more than 1 change that needs to come into this branch, impacting this file.  

Do we need to go back to apply this to 6.0.2?

randyh62:
> > I had to update the contents of the using-gpu-sanitizer.md file more than just the cherry-pick was providing.
> 
> That usually means there are more than 1 change that needs to come into this branch, impacting this file.
> 
> Do we need to go back to apply this to 6.0.2?

The request was to apply the change all the way back to 6.0.0 as it should have been documented from the start. But I will close this and check.

randyh62:
It turns out the changes are needed here.

peterjunpark (changes requested):
Linting error

peterjunpark:
This fix shouldn't be updating the `.wordlist.txt`. That doesn't make sense because your other PRs weren't flagged

randyh62:
> This fix shouldn't be updating the `.wordlist.txt`. That doesn't make sense because your other PRs weren't flagged
It needed the addition of LSan, which apparently was already in the other wordlists.

peterjunpark:
I don't see it in 6.0.0 https://github.com/ROCm/ROCm/blob/docs/6.0.0/.wordlist.txt
https://github.com/ROCm/ROCm/pull/3292/files

randyh62:
> I don't see it in 6.0.0 https://github.com/ROCm/ROCm/blob/docs/6.0.0/.wordlist.txt https://github.com/ROCm/ROCm/pull/3292/files

And yet it uses LSan. But the only test was the readthedocs build.

peterjunpark (approved):

PR #3332, opened 2024-06-19
Add RHEL 9.4 to compat matrix

peterjunpark:


yhuiYH (approved):
LGTM too.  We should fix those # for 6.0 and 6.1 to give them actual reference names.  But we can fix those later.

PR #3358, opened 2024-06-25
Remove article info for moved or deleted pages

samjwu:
These pages no longer exist or were moved to install docs


yhuiYH (approved):
LGTM

PR #3362, opened 2024-06-26
Fix card text color in index

peterjunpark:
This PR removes unnecessary the `sd-text-black` class from cards on the home page.
- This lets the text colour in those cards switch between black and white in light mode and dark mode, respectively.


randyh62 (approved):

yhuiYH (approved):
Wow, nice, you found it!

samjwu (approved):

PR #3363, opened 2024-06-26
ROCm landing page reorg

randyh62:
edit index.md to reorg tiles and update Intro
edit _toc.yml.in to reflect new order
add Concepts banner image


docs/index.md
```diff
@@ -5,61 +5,56 @@
...
 
 ::::{grid} 1 2 2 2
 :class-container: rocm-doc-grid
 
 :::{grid-item-card}
 :class-card: sd-text-black
-:img-top: ./data/banner-installation.jpg
-:img-alt: Install documentation
+:img-top: ./data/banner-concepts.jpg
+:img-alt: Concepts documentation
```
yhuiYH, line 30:
```suggestion
:img-alt: Explanation documentation
```
Do we really want to change from Conceptual -> Concepts?
Maybe we should just drop back to using "Explanation" instead, like the original quadrant:
https://diataxis.fr/

randyh62:
Moved to 6.2

PR #3373, opened 2024-06-28
External CI: Fix a typo from composable_kernel pipeline

alexxu-amd:
The flag should be INSTANCES_ONLY instead of INSTANCE_ONLY.
Also added libdrm-dev to suppress warnings


abhimeda (approved):
Looks good!

PR #3378, opened 2024-06-29
Update single-gpu-fine-tuning-and-inference.rst with correct `amd-smi` flag

james-banks:
Updated to correctly use the `--showproductname` flag.

Prior flag of `-showproductname` was not valid.


peterjunpark (approved):
Hi @james-banks,

Thanks so much for pointing this out. LGTM

PR #3390, opened 2024-07-03
External CI: Add symlink to lib/llvm folder for ROCmValidationSuite

alexxu-amd:
Lately, /rocm/llvm has been moved to /rocm/lib/llvm, which breaks ROCmValidationSuite pipeline. As a temporary solution, create a symlink to /rocm/lib/llvm folder

Successful build with latest llvm-project: https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=3058&view=logs&j=a9dd7315-711f-5623-6d52-e9ce616061c5

In the long term, the install script of llvm-project should be change to /rocm/lib/llvm and we need to modify all pipelines' cmake flags to adapt that.

### DIFF
--- a/.azuredevops/components/ROCmValidationSuite.yml
+++ b/.azuredevops/components/ROCmValidationSuite.yml
@@ -13,7 +13,7 @@ parameters:
     - libyaml-cpp-dev
     - libpci-dev
     - libpci3
-    - googletest
+    - libgtest-dev
     - git
 - name: rocmDependencies
   type: object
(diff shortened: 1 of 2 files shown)


abhimeda (approved):

.azuredevops/components/ROCmValidationSuite.yml
```diff
@@ -59,10 +63,17 @@ jobs:
       parameters:
         dependencyList: ${{ parameters.rocmDependencies }}
         dependencySource: tag-builds
+# Set link to redirect llvm folder
+  - task: Bash@3
+    displayName: update cmake
```
danielsu-amd, line 68: Can this be changed to "create symlink" or similar

alexxu-amd: good catch

danielsu-amd (approved):

PR #3395, opened 2024-07-04
Initial draft of MI300A tuning guide, with images, toc, and index page changes

amd-jnovotny:
This guide incorporates content from the draft document along with supporting infrastructure changes to fill out the document, which I based on the MI200 guide. The merge will be coordinated with Peter Park, who is reworking the tuning-guide landing page.


robers97:
I'm going to put my review in here in the form of 
Current --> recommended update
I expect the recommendations be considered in the context and not necessarily adopted verbatim.  I'll make them accurate, but embellishment to better bridge to a 'user' would be beneficial.

These settings are critical for performance because the OS is now responsible for memory management across the CPU and GPU accelerators.  --> These settings are critical for performance because the OS on an APU is responsible for memory management across the CPU and GPU accelerators.

Merge System Settings paragraph 2 and 3.

In addition, some settings limit GPU memory allocation. --> In the APU memory model, system settings are available to limit GPU memory allocation.

The System BIOS settings section has 1 sentence fragment.  This section is a carryover from the previous document.  ROCM users are presumably going from one generation to the next.  The APU model is different so while there is not much to do in this section, some embellishment would be beneficial.

Will start again after a more critical review is done.  My original input was just notes... I'm from the system team and there doesn't appear to be continuity from anyone closer to the ROCM user base.

amd-jnovotny:
Closing and moving to a new branch

PR #3412, opened 2024-07-11
External CI: Change composable_kernel pipeline to build for specific GPUs with tests and examples

alexxu-amd:
We previously had some errors when trying to build for multiple architectures with tests and examples. So the work around was to drop tests and examples by setting INSTANCES_ONLY=ON and build for all GPUs. Now the error has been fixed by https://github.com/ROCm/composable_kernel/commit/959073842c0db839d45d565eb260fd018c996ce4. Change CK back to build for specific GPUs with tests and examples, this will reduce the build time to ~1.5 hrs.

Successful CK build for gfx1100;gfx1030;gfx941:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=3347&view=results

Successful hipTensor (uses CK as dep) build for gfx941:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=3376&view=results

Also, changing all gfx90a targets to gfx942 since we got MI300 instead of MI250.


danielsu-amd (approved):
Looks good

PR #3457, opened 2024-07-24
Remove custom template for mivisionx changelog from autotag

samjwu:
Closes https://github.com/ROCm/ROCm/issues/3430

The custom template for MIVisionX is no longer needed as the base template correctly scrapes it now


yhuiYH (approved):
LGTM

PR #3476, opened 2024-07-30
External CI Build Job for Pytorch WHL

amd-jmacaran:
Passing build job (non-scheduled version) can be found here
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=4458&view=results

Produces a 1677 MB wheel file for gfx942 that includes latest default branch builds of ROCm software stack and the latest default branch build of upstream pytorch.

Will create versions for other combinations of GPU target, ROCm branch, and upstream pytorch branch.

### DIFF
--- a/.azuredevops/nightly/pytorch.yml
+++ b/.azuredevops/nightly/pytorch.yml
@@ -145,7 +145,7 @@ jobs:
     value: -GNinja
   - name: DESIRED_DEVTOOLSET
     value: cxx11-abi
-  pool: rocm-ci_larger_base_disk_pool
+  pool: ${{ variables.ULTRA_BUILD_POOL }}
   workspace:
     clean: all
   steps:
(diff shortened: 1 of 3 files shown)


danielsu-amd (dismissed):
I was able to install the wheel and successfully run the following:
```bash
python3 -c 'import torch' 2> /dev/null && echo 'Success' || echo 'Failure'
python3 -c 'import torch; print(torch.cuda.is_available())'
python3 -c 'import torch; x = torch.rand(5, 3); print(x)'
# https://pytorch.org/get-started/locally/#linux-verification
# https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/3rd-party/pytorch-install.html#testing-the-pytorch-installation
```

Tried running the tests at https://github.com/pytorch/builder/blob/main/run_tests.sh, but they are all disabled anyways.

danielsu-amd (approved):

PR #3525, opened 2024-08-06
Use intersphinx link to rocr debug agent

amd-jnovotny:
change from absolute link to intersphinx link


samjwu (approved):

amd-jnovotny:
The linting errors are a result of new spellchecks being added to the templates. They aren't a result of my PR.

PR #3558, opened 2024-08-09
Added Docker build script and fixed bare metal build script

ammallya:


docker_run.sh
```diff
@@ -0,0 +1,71 @@
...
+    -e DOCK_WORK_FOLD=/src \
+    -w /src \
+    -v $PWD:/src \
+    --mount="type=bind,src=${fake_group},dst=/etc/group,readonly" \
+    --mount="type=bind,src=${fake_passwd},dst=/etc/passwd,readonly" \
+    -v ${HOME}/.ccache:${HOME}/.ccache \
+    -u $(id -u):$(id -g) \
+    ${BASE_IMAGE} bash
+
+rm "${fake_group}" "${fake_passwd}"
```
amd-isparry, line 71: This is a bad idea. It disguises the return code from docker, so at the end you only know if it managed to gelete the fake group and password files, not if the commands that were run in docker worked or not.

At the very least the return code from docker should be captured and used as the return code of the script.

amd-isparry (changes requested):

PR #3599, opened 2024-08-16
css tests for Text over background image

randyh62:
Test PR...


yhuiYH:
I updated the graphic and adjusted padding/font, to make it look more like what we have:
https://github.com/ROCm/rocm-docs-core/pull/915

PR #3618, opened 2024-08-20
External CI: Pipeline dependency adjustments for hipBLAS-common

amd-jmacaran:


.azuredevops/components/hipBLASLt.yml
```diff
@@ -23,7 +23,6 @@ parameters:
   type: object
   default:
     - clr
-    - hipBLAS
```
alexxu-amd, line 26: We don't need to remove hipBLAS anymore. For some reason, the error no longer exists after merging the PR into dev. It works with hipBLAS now.

alexxu-amd: https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=5779&view=logs&j=a1cf3924-7251-5656-99ce-ba94dfad1583&t=fedc02f7-3fa4-5d67-87bf-0cf28b8351b3

amd-jmacaran: It's not a matter of need, but requested by Brad. It's also good practice to keep dependency list accurate for keeping map of dependencies maintained.

alexxu-amd: Gotcha, I am surprised hipBLASLt no longer depends on hipBLAS. I was worried about removing hipBLAS would introduce an error but since it's requested by Brad, then we should remove it

alexxu-amd (approved):

PR #3627, opened 2024-08-21
fixed changes mentioned in review

ammallya:


WBobby (commented):
After running docker_run.sh and enter the docker container, I see:
oem@4d1f26187870:/src$ sudo apt update
sudo: account validation failure, is your account locked?
sudo: a password is required
Please double check and set shadow file correctly.

ammallya:
Changes made to code

tools/rocm-build/docker/ubuntu20/docker_run.sh
```diff
@@ -0,0 +1,40 @@
...
+    -e CCACHE_ENABLED=true \
+    -e DOCK_WORK_FOLD=/src \
+    -w /src \
+    -v $PWD:/src \
+    --mount="type=bind,src=${fake_passwd},dst=/etc/passwd,readonly" \
+    --mount="type=bind,src=${fake_shadow},dst=/etc/shadow,readonly" \
+    --mount="type=bind,src=${fake_group},dst=/etc/group,readonly" \
+    -v ${HOME}/.ccache:/.ccache \
+    -u $(id -u):$(id -g) \
+    --cap-add SYS_ADMIN \
```
WBobby, line 34: Do not add extra permissions.

tools/rocm-build/docker/ubuntu20/docker_run.sh
```diff
@@ -0,0 +1,40 @@
...
+    --cap-add DAC_READ_SEARCH \
```
WBobby, line 35: Same as above.

WBobby (commented):
Please run a test ROCm build to check if the scripts works correct or not.

tools/rocm-build/docker/ubuntu20/docker_run.sh
amd-isparry: I totally agree. 
If you want flexability consider adding additional options from decoding command line options.

tools/rocm-build/docker/ubuntu20/docker_run.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 1
+fi
+
+fake_passwd="$PWD/passwd"
+fake_shadow="$PWD/shadow"
+fake_group="$PWD/group"
+
+sudo getent passwd > "${fake_passwd}"
+sudo getent group > "${fake_group}"
+sudo getent shadow > "${fake_shadow}"
```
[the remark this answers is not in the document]
ammallya: I tried that but it kept giving me auth errors.

sudo: account validation failure, is your account locked?
sudo: a password is required

which Bobby pointed out

tools/rocm-build/docker/ubuntu22/docker_run.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ammallya: Aakash told that he wanted to proceed this way so theres a script in each os version folder. I will discuss with him today figure out a solution.

tools/rocm-build/docker/ubuntu20/docker_run.sh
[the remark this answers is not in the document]
amd-isparry: I don't care how you do it, as long as you don't use "sudo".
Using 'sed' is the approach I would use, but if you want to do it differently then that is fine.

ammallya:
Fixed the code and security loophole issues

tools/rocm-build/docker/ubuntu20/docker_run.sh
```diff
@@ -30,10 +33,14 @@ docker run -ti \
     --mount="type=bind,src=${fake_shadow},dst=/etc/shadow,readonly" \
     --mount="type=bind,src=${fake_group},dst=/etc/group,readonly" \
     -v ${HOME}/.ccache:/.ccache \
+    -v ${HOME}:/home/${CURRENT_USER} \
```
amd-isparry: No, this is a really really bad thing to do. It means that the user will have their home directory mounted into the build environment, so all their aliases and configuration will be used.

This breaks the idea that the build is the same for everyone. Consider a person who changes their PATH to include ~/bin, and then has their own ~/bin/ln command. The point of using a container to provide the build environment is to make it easy to set up the reproducible build environment.

There are other ways we could set up the build environment, for example we could ship a tarball of a chrooted setup, and tell people to use chroot to enter this subtree. We could use VMs. We could use podman. We are using docker containers for ease of use, but as you can see they are not the only choice. These days this idea seems to be called devcontainers.

We could mount a tmpfs at the $HOME location inside the container. We can also set HOME inside the container to a fixed location. The latter helps with reproducible builds.

amd-isparry (changes requested):

tools/rocm-build/docker/ubuntu20/docker_run.sh
ammallya: I see I did it this way because while using the make command the script was failing because it could not find /home/user 
maybe I can just make a directory with the name of the current user instead of mounting it.

ammallya:
Fixed Home mount issue raised.

amd-isparry:
[AMD Official Use Only - AMD Internal Distribution Only]

Let me expand a bit on the reproducible builds. It would be very nice if we could have two people, say in India and California, both starting a build on their local machines and end up with two sets of packages which are identical. There are lots of things that work against this, but we are tackling them. Obviously we want to start with the same source code. We get this by using the “repo” tool and a manifest which gets us the same files checked out in the same locations. We have a version controlled build environment, so we know we are using the same tools (compilers, shells, grep, make and so on). We know how to set timestamps embedded in files to uniform values, rather than the local time so this takes care of building in different time zones at different times. We would like any generated files whose names end up embedded in the product to be the same. This suggests that we should set HOME to a known value, and maybe LOGNAME and USER as well.

Sparry, Icarus
PMTS Software Development Eng. | AMD
ML PMO&Validation US
D 8342106<tel:8342106>

PR #3631, opened 2024-08-21
External CI: Support for rocPyDecode

amd-jmacaran:
Passing build log: https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=5957&view=results


danielsu-amd (approved):
The hip-python package is interesting, we might have to address that if this ever gets more pip dependencies.

PR #3648, opened 2024-08-27
Correct ttm to amdttm

ckime-amd:
The original content incorrectly had `ttm` in path rather than `amdttm`.


ckime-amd:
```
chrikime@x1001c0s0b0n0:~> cat /sys/module/ttm/parameters/pages_limit
cat: /sys/module/ttm/parameters/pages_limit: No such file or directory
chrikime@x1001c0s0b0n0:~> cat /sys/module/amdttm/parameters/pages_limit
98540199
```

amd-jnovotny (approved):
looks good.

PR #3653, opened 2024-08-28
Expand the section on changing thread affinity

amd-jnovotny:
This incorporates a change to the thread affinity section. More changes still to be added.

### DIFF
--- a/docs/how-to/system-optimization/mi300a.rst
+++ b/docs/how-to/system-optimization/mi300a.rst
@@ -242,8 +242,11 @@ This section describes performance-based settings.
   
   This change prevents internal ROCm threads from having their CPU core affinity mask 
   set to all CPU cores available. With this setting, the threads inherit their parent's 
-  CPU core affinity mask. If you have any questions regarding this setting, 
-  contact your MI300A platform vendor. To enable this setting, enter the following command:
+  CPU core affinity mask. Before adjusting this setting, ensure you thoroughly understand 
+  your system topology and how the application, runtime environment, and batch system
+  set the thread-to-core affinity. If you have any questions regarding this setting, 
+  contact your MI300A platform vendor or the AMD support team. 
+  To enable this setting, enter the following command:
 
   .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


lpaoletti (approved):

docs/how-to/system-optimization/mi300a.rst
ckime-amd, line 1: Looking to make some additional changes around 148-150:
> These settings can also be hardcoded in the ``/etc/modprobe.d/amdttm.conf`` file. To use this method, 
>  the filesystem must already be set up when the kernel driver is loaded.
>  Add the following lines to ``/etc/modprobe.d/amdttm.conf``:

Not sure how best to highlight, but, along with " the filesystem must already be set up when the kernel driver is loaded", need to note something like: 

> If the filesystem is not already setup when the kernel driver is loaded, then the options must be specified as boot parameters when the kernel is loaded.  For example:
> ```
> vmlinuz-[...] amdttm.pages_limit=134217728 amdttm.page_pool_size=134217728 [...]
> ```

ckime-amd (changes requested):
Requesting the additional text to note that parameters need to be added as kernel parameters if the filesystem is not available.  Feel free to DM me if we need to discuss, I will do my best to answer any questions.

amd-jnovotny:
@ckime-amd I've incorporated your suggestions into the draft. Let me know if anything else needs to be added.

docs/how-to/system-optimization/mi300a.rst
```diff
@@ -145,15 +145,26 @@ This section describes performance-based settings.
...
      options amdttm page_pool_size=134217728
 
+  If the filesystem is not already set up when the kernel driver is loaded, then the options
+  must be specified as boot parameters. To specify the settings
+  as boot parameters when loading the kernel, use this example as a guideline:
+
+  .. code-block:: shell
+
+     vmlinux-[...] amdttm.pages_limit=134217728 amdttm.page_pool_size=134217728 [...]
+
```
ckime-amd, line 167: Thanks, this looks good to me.  We could get more eyes on it, but this would only introduce further delay and they missed these details earlier regardless.  We'll look for feedback from users of the documentation to guide us if further improvements can be made.

ckime-amd (approved):
Thanks for the additional text.  Looks good, and I like the editorial choices.  Approving.

lpaoletti (approved):

PR #3668, opened 2024-09-03
Fix intersphinx links

peterjunpark:
This PR fixes intersphinx links in ROCm/ROCm based on Sphinx warnings/errors.

The dead link checker cannot find broken intersphinx links. Broken intersphinx links must be found as part of the review process by checking the Read the Docs build trace or by using `linkcheck` (https://docs.readthedocs.io/en/stable/guides/intersphinx.html).


yhuiYH (approved):
LGTM.  Argh, and the redirects don't work on intersphinx links.  Doh, I moved so many files around in 6.2.0 :(

PR #3728, opened 2024-09-16
Update link to rocRAND data type support

amd-jnovotny:
The change in the link target is required due to changes in https://github.com/ROCm/rocRAND/pull/548.


samjwu (approved):
approved, depending on merging https://github.com/ROCm/rocRAND/pull/548 before this one

PR #3730, opened 2024-09-16
Add starting-versions flag to autotag script

samjwu:
Closes https://github.com/ROCm/ROCm/issues/3585

ie: add a prev release option to autotag script so that it does not have to start at 5.0.0

# Before

passing in the `--do-previous` flag would start from 5.0.0

# After

pass in the `starting-versions=<a.b.c>` flag where a.b.c is the version to start from


yhuiYH (approved):
LGTM

PR #3735, opened 2024-09-17
External CI: rdc smoke tests

amd-jmacaran:
- Referred to public documentation, build instructions, source code in tests directory, and iterative runs to modify build flags.
- rdci test failures are known due to singleton nature of rocprofiler, but gtest attempting to spawn multiple instances. There is an internal ticket to track the issue.

[Build Log with the passing rdcd and known failing rdci tests](https://dev.azure.com/rocm-ci/ROCm-CI/_build/results?buildId=8065&view=results)

### DIFF
--- a/.azuredevops/components/rdc.yml
+++ b/.azuredevops/components/rdc.yml
@@ -18,16 +18,35 @@ parameters:
     - autoconf
     - libtool
     - pkg-config
+    - libdrm-dev
+    - libyaml-cpp-dev
 - name: rocmDependencies
   type: object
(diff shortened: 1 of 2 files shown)


amd-jmacaran:
> The test job shows as passing even though the rdci tests are failing, is that something we can fix?

That's a product of the rdctst's exit/return code still being 0, even with the failure. One future task is to convert tests like these that aren't automatically parsed and graphed to a format that is parsed and graphed. That parser can mark failures accordingly.

PR #3824, opened 2024-09-27
External CI: add rocprofiler-sdk pipeline IDs

danielsu-amd:


alexxu-amd:
Any reason why we changed rocprofile-register from amd-mainline to amd-staging? I saw their default branch is still amd-mainline

danielsu-amd:
Yea, for both rocprofiler-register and sdk their default Github branch is amd-mainline, but the latest source is still on amd-staging. I put amd-staging so the CI is pulling that latest source like for other components, but you're right it's not actually the default branch. @amd-jmacaran what do you think?

See:
https://github.com/ROCm/rocprofiler-sdk/tree/amd-mainline
https://github.com/ROCm/rocprofiler-sdk/tree/amd-staging (3 commits ahead)

amd-jmacaran (approved):
Default branch for External CI use case should be amd-staging for these components. I still have that draft PR to support amd-mainline builds.

PR #3843, opened 2024-10-01
added revised note re. radeon gpus (#3839)

spolifroni-amd:


spolifroni-amd:
There may be changes to this. Closing for now

PR #3851, opened 2024-10-01
Add a page explaining IOMMU to Conceptual section

alexxu-amd:


docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Overview
+================================================================
+
+The IOMMU (IO Memory Management Unit) is a device which provides memory remapping services to 
+IO devices on a platform. Similar to the MMU on a CPU, the IOMMU is able to provide a private 
+virtual address space for each device or group of devices on a platform. While the IOMMU is 
+most commonly used for virtualization provide a private IO space for devices used in a guest 
+virtual machine, it can also be used to isolate devices on bare metal to prevent accidental
+or nefarious DMA accesses to things that they should should not have access to. The IOMMU provides
```
spolifroni-amd, line 17: What things exactly?

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+virtualized access to the entire platform address space, not just system memory. It can also
+provide virtualized access to the MMIO portions of the platform address space. This is used for
+things like peer to peer DMA. In addition to DMA remapping, the IOMMU also support interrupt remapping.
+In addition to virtualization use cases, this can be advantageous for devices which support multiple
+interrupts on bare metal and it may be required for interrupt delivery on platforms with lots of cores.
+
```
spolifroni-amd, line 23: I only see Peer to Peer DMA in the topics. Is that what we want to emphasize?

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+================================================================
+
+Peer to Peer DMA over PCIe
+----------------------------------------------------------------
+
+Peer to Peer DMA is not officially part of the PCI/PCIe spec. As such, the behavior for peer to peer DMA
+varies from platform to platform. AMD platforms prior to zen only supported peer to peer DMA for writes.
+On Zen and newer CPUs, peer to peer DMA is fully supported. On Intel, peer to peer DMA is supported with
+a number of limitations on various platforms. For example only certain root ports on the platform support it.
+If one of your devices happens to be on an unsupported root port, it will not work. The Linux kernel has
```
spolifroni-amd, line 35: how does a user check if this is supported?

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+special peer to peer DMA code to handle these cases. It provides an API for drivers to use which can look
+at two devices and determine whether or not they can support peer to peer DMA. This functionality is controlled
+by the ``CONFIG_PCI_P2PDMA`` kernel config option. In addition to that parameter, you also need to enable
+``dma-buf`` (the framework for sharing DMA memory across drivers). To properly support peer to peer DMA with the IOMMU,
+the following options must be enabled in your kernel config: ``CONFIG_PCI_P2PDMA``, ``CONFIG_DMABUF_MOVE_NOTIFY`` 
+and ``CONFIG_HSA_AMD_P2P``.
+
```
spolifroni-amd, line 42: is there some linux doc that can be pointed to?

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+  All devices on Linux register their DMA addressing capabilities (e.g. 32 bit, 40 bit, 44, bit, etc. i.e., how much address
+  space can they natively access). The kernel will then make sure that any address space mapped for DMA is mapped within
+  the DMA addressing limits of the device. Additionally, only address space explicitly mapped by the device is mapped into
+  its virtual address space so an access to an unmapped page will generate an IOMMU page fault. This is generally the
+  default when the IOMMU is enabled.
+
+Advantages and disadvantages
+================================================================
+
+There advantages and disadvantages depending on the devices and platforms you are using.
```
spolifroni-amd, line 76: Advantages and disadvantages to what? To using an IOMMU?

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Peer to Peer DMA
+================================================================
+
+Peer to Peer DMA over PCIe
+----------------------------------------------------------------
+
+Peer to Peer DMA is not officially part of the PCI/PCIe spec. As such, the behavior for peer to peer DMA
+varies from platform to platform. AMD platforms prior to zen only supported peer to peer DMA for writes.
+On Zen and newer CPUs, peer to peer DMA is fully supported. On Intel, peer to peer DMA is supported with
```
spolifroni-amd, line 33: Is the behaviour on Intel impotrtant here?

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,98 @@
...
+****************************************************************
+Input-Output Memory Management Unit (IOMMU)
+****************************************************************
+
+Overview
+================================================================
+
+The IOMMU (IO Memory Management Unit) is a device which provides memory remapping services to 
+IO devices on a platform. Similar to the MMU on a CPU, the IOMMU is able to provide a private 
+virtual address space for each device or group of devices on a platform. While the IOMMU is
```
spolifroni-amd, line 14: By "platform", do we mean operating systems, or an entire OS-GPU config?

docs/conceptual/iommu.rst
spolifroni-amd: Is it that the IOMMU can be enabled/disabled?

alexxu-amd: Yup, IOMMU has 3 modes, disabled, passthrough mode and remapping mode, where the latter two are considered enabled. I believe the advantage and disadvantage are regarding to enable/disable IOMMU.

docs/conceptual/iommu.rst
alexxu-amd: I believe things refer to the protected memory space

docs/conceptual/iommu.rst
alexdeucher: IOMMU generally provides IO remapping services for all devices in a system (aka platform).

docs/conceptual/iommu.rst
alexdeucher: replace "things" with "system addresses (memory or MMOIO space allocated to other devices or users"

docs/conceptual/iommu.rst
alexdeucher: The IOMMU is required to support multiple interrupts per device on devices which support MSIs in Linux.  Supporting multiple interrupts per device without the IOMMU requires MSI-X which not all devices support.

docs/conceptual/iommu.rst
alexdeucher: Maybe say "AMD Zen" rather than just "Zen".  As for Intel, I thought it was useful to call out the limitations since customers may be using Intel platforms.

docs/conceptual/iommu.rst
spolifroni-amd: But is that something we want to emphasize? Do we want a separate topic for interrupt remapping?

docs/conceptual/iommu.rst
spolifroni-amd: So since this is AMD documentation, can this be omitted?

docs/conceptual/iommu.rst
spolifroni-amd: That's not really answering my question.

When the term "platform" is being used here, what is it referring to? I need specifics so I can be as clear as possible in the text.

docs/conceptual/iommu.rst
alexdeucher: s/MMOIO/MMIO/

docs/conceptual/iommu.rst
alexdeucher: It provides context as to why you may want some level of IOMMU enabled.  E.g., we may have users that have a system with a lot of cores and some NICs that need multiple MSIs to perform optimally.  These requirements would require IOMMU, but at the same time, they don't want the extra overhead of the IOMMU for P2P DMA between GPUs.  Setting iommu=pt would provide a middle ground by allowing the interrupt remapping, but not using device IO remapping.

docs/conceptual/iommu.rst
alexdeucher: It was just as context to provide some details of the platform specific limitations that a user might run into.  E.g., if a user tries to use P2P DMA on an Intel system and it doesn't work between some of the root ports, this is why.  I.e., this is a platform limitation rather than a driver bug.  If you think it's too confusing, feel free to omit it.

docs/conceptual/iommu.rst
alexdeucher: platform = the system.  The motherboard and CPU and all of the devices connected to it.  If you take a server with a Zen CPU in it, the IOMMU integrated into the CPU provides IO and interrupt remapping services to all of the devices on the system (PCIe devices, USB, SATA, M.2, etc.).

docs/conceptual/iommu.rst
alexdeucher: I think the point is to provide some context around IOMMU in general.  A lot of users don't know what it is and there is a lot of conflicting information about it.  For various historical reasons, AMD had recommended that users disable it for either compatibility or performance reasons for ROCm.  Then as the number of cores on our CPUs increased we had to temper our statement to make sure systems with a lot of cores worked properly.  So we recommend iommu=pt rather than iommu=off.  As such, the level of validation we do with the IOMMU fully enabled has always been more limited.  In the long term however, it's an important compatibility and security tool that we should embrace and encourage.  It protects devices from buggy DMAs from other devices and it allows for P2P DMA on systems with large numbers of devices which may not be possible with the system address space layouts on most modern systems.  So now we are at another crossroads where we need to reconsider our recommended stance on IOMMU settings.

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
+.. meta::
+   :description: Input-Output Memory Management Unit (IOMMU)
+   :keywords: IOMMU, DMA, PCIe, xGMI, AMD, ROCm
+
+****************************************************************
+Input-Output Memory Management Unit (IOMMU)
+****************************************************************
+
+The I/O Memory Management Unit (IOMMU) provides memory remapping services for I/O devices. It adds support for address translation and system memory access protection on direct memory access (DMA) transfers from peripheral devices.
```
lpaoletti, line 9: Instead of "It adds support" should it be something like "It supports address translation..."

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+   :description: Input-Output Memory Management Unit (IOMMU)
+   :keywords: IOMMU, DMA, PCIe, xGMI, AMD, ROCm
+
+****************************************************************
+Input-Output Memory Management Unit (IOMMU)
+****************************************************************
+
+The I/O Memory Management Unit (IOMMU) provides memory remapping services for I/O devices. It adds support for address translation and system memory access protection on direct memory access (DMA) transfers from peripheral devices. 
+
+The IOMMU's memory remapping services are used to:
```
lpaoletti, line 11:
```suggestion
The IOMMU's memory remapping services:
```

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+* provide private I/O space for devices used in a guest virtual machine.
+* prevent unauthorized DMA requests to system memory and to memory-mapped I/O (MMIO).
+* help in debugging memory access issues.
+* facilitate peer-to-peer DMA.
+
+The IOMMU also provides interrupt remapping, which is used by devices that support multiple interrupts and for interrupt delivery on hardware platforms with a large number of cores.
+
+.. note::
+
+  AMD Instinct accelerators are connected via XGMI links and don't use PCI/PCIe for peer-to-peer DMA. Because PCI/PCIe is not used for peer-to-peer DMA, there are no device physical addressing limitations or platform root port limitations. However, because non-GPU devices such as RDMA NICs use PCIe for peer-to-peer DMA, there might still be physical addressing limitations and platform root port limitations when these non-GPU devices interact with other devices, including GPUs.
```
lpaoletti, line 22:
```suggestion
  AMD Instinct accelerators are connected via XGMI links and don't use PCI/PCIe for peer-to-peer DMA. Because PCI/PCIe is not used for peer-to-peer DMA, there are no device physical addressing limitations or platform root port limitations. However, because non-GPU devices such as RDMA NICs use PCIe for peer-to-peer DMA, there might still be physical addressing and platform root port limitations when these non-GPU devices interact with other devices, including GPUs.
```

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Linux supports IOMMU in both virtualized environments and bare metal. 
+
+The IOMMU is enabled by default but can be disabled or put into passthrough mode through the Linux kernel command line:
```
lpaoletti, line 26:
```suggestion
The IOMMU is enabled by default but can be disabled or put into passthrough mode through a Linux kernel command line:
```

lpaoletti: Not sure about that edit

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+  :widths: 25 25 50
+  :header-rows: 1
+
+  * - IOMMU Mode
+    - Kernel command
+    - Description
+  * - Enabled
+    - default setting
+    - The IOMMU is enabled in remapping mode. Each device gets its own I/O virtual address space. All devices on Linux register their DMA addressing capabilities and the kernel will ensure that any address space mapped for DMA is mapped within the device's DMA addressing limits. Only address space explicitly mapped by the devices will be mapped into virtual address space. Attempts to access an unmapped page will generate an IOMMU page fault. This setting is recommended for AMD Radeon GPUs that need peer-to-peer DMA.
+  * - Passthrough
```
lpaoletti, line 38:
```suggestion
    - The IOMMU is enabled in remapping mode. Each device gets its own I/O virtual address space. All devices on Linux register their DMA addressing capabilities, and the kernel will ensure that any address space mapped for DMA is mapped within the device's DMA addressing limits. Only address space explicitly mapped by the devices will be mapped into virtual address space. Attempts to access an unmapped page will generate an IOMMU page fault. This setting is recommended for AMD Radeon GPUs that need peer-to-peer DMA.
  * - Passthrough
```

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+    - ``iommu=pt``
+    - The IOMMU is enabled with interrupt remapping enabled but I/O remapping disabled. The entire platform shares a common platform address space for system memory and MMIO spaces, ensuring compatibility with drivers from external vendors, while still supporting CPUs with a large number of cores. This setting is recommended for AMD Instinct Accelerators and for AMD Radeon GPUs that don't need peer-to-peer DMA.
```
lpaoletti, line 40: The first sentence seems off.

The IOMMU is enabled with interrupt remapping enabled and the I/O remapping is disabled. ???

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+  * - Disabled
+    - ``iommu=off``
+    - The IOMMU is disabled and the entire platform shares a common platform address space for system memory and MMIO spaces.
+
+The IOMMU also provides virtualized access to the MMIO portions of the platform address space for peer-to-peer DMA.
+
+Because peer-to-peer DMA is not officially part of the PCI/PCIe specification, the behavior of peer-to-peer DMA varies between hardware platforms. 
+
+AMD CPUs earlier than AMD Zen only supported peer-to-peer DMA for writes. On CPUs from AMD Zen on, peer-to-peer DMA is fully supported.
```
lpaoletti, line 49:
```suggestion
AMD CPUs earlier than AMD Zen only supported peer-to-peer DMA for writes. On CPUs from AMD Zen and later, peer-to-peer DMA is fully supported. 
```

docs/conceptual/iommu.rst
```diff
@@ -0,0 +1,55 @@
...
+    - ``iommu=off``
+    - The IOMMU is disabled and the entire platform shares a common platform address space for system memory and MMIO spaces.
+
+The IOMMU also provides virtualized access to the MMIO portions of the platform address space for peer-to-peer DMA.
+
+Because peer-to-peer DMA is not officially part of the PCI/PCIe specification, the behavior of peer-to-peer DMA varies between hardware platforms. 
+
+AMD CPUs earlier than AMD Zen only supported peer-to-peer DMA for writes. On CPUs from AMD Zen on, peer-to-peer DMA is fully supported. 
+
+To use peer-to-peer DMA on Linux, the following options must be enabled in your Linux kernel configuration:
```
lpaoletti, line 51:
```suggestion
To use peer-to-peer DMA on Linux, enable the following options in your Linux kernel configuration:
```

lpaoletti (approved):

Issue #3743, opened 2024-09-18
[Issue]:  brocken cp-command in half post-build script


jdumke:
### Problem Description

While trying to move package files, after building half, the cp command brakes in cause of incomplete filename-pattern.
I can't locate where the cp call takes place to fix it.

### Operating System

Debian 12 Bookworm with Backports

### CPU

 AMD Ryzen 5 5600G with Radeon Graphics

### GPU

AMD Radeon Pro W7900

### ROCm Version

ROCm 6.2.0

### ROCm Component

half

### Steps to Reproduce

Like described in README.md in ROCm.git 
local settings: export GPU_ARCHS="gfx803;gfx90c"

### Additional Information

[all fine until …]
-- Install configuration: "Release"
-- Installing: /opt/rocm-6.2.0/include/half/half.hpp
-- Installing: /opt/rocm-6.2.0/share/doc/half/LICENSE.txt
gmake[1]: Leaving directory '/home/builder/devel/amd/rocm-6.2/out/debian-12/12/build/half'
+ rm -rf _CPack_Packages/
+ find -name '*.o' -delete
+ mkdir -p /home/builder/devel/amd/rocm-6.2/out/debian-12/12//half
+ cp '/home/builder/devel/amd/rocm-6.2/out/debian-12/12/build/half/*.' /home/builder/devel/amd/rocm-6.2/out/debian-12/12//half
cp: cannot stat '/home/builder/devel/amd/rocm-6.2/out/debian-12/12/build/half/*.': No such file or directory

real	0m2.276s
user	0m0.713s
sys	0m0.290s
+ mv /home/builder/devel/amd/rocm-6.2/out/debian-12/12/logs/half.inprogress /home/builder/devel/amd/rocm-6.2/out/debian-12/12/logs/half.errors
+ echo Error in half
Error in half
+ exit 1

schung-amd:
Hi @jdumke, not sure if this is causing your issue, but at first glance I notice you're setting `export GPU_ARCHS="gfx803;gfx90c"`. If you're on a W7900, the correct setting is `export GPU_ARCHS="gfx1100"`. Also, you should disable your integrated graphics, as this can cause some issues. I'll try to repro your issue, but hopefully this helps.

jdumke:
I don't build for a W7900, I'm building for a W5100 instead, but I must choose a card from the drop down to open the issue and "Legacy" is not an option.

schung-amd:
[...]

PR #3859, opened 2024-10-03
Error handling for unsupported distros

schung-amd:
Currently on unsupported distros envsetup.sh does not define a package type, and the build scripts fail far downstream. Adds error handling for this before progressing with build. Addresses unclear logs as seen in issue #3743.


schung-amd:
Does this need to be ported internally?

schung-amd:
Stale, usecase for this probably obsoleted by TheRock.

PR #3898, opened 2024-10-15
Update bitsandbytes default branch in docs

peterjunpark:
Users should checkout `rocm_enabled_multi_backend` by default instead of `rocm_enabled` now.

https://github.com/ROCm/bitsandbytes


amd-jnovotny (approved):
looks good

PR #3936, opened 2024-10-23
Improve consistency of the gpu-arch-specs table.

bcornille:
There was inconsistency in the CU count where 250X lists per GCD, but 250 does not. In keeping with the per GCD convention, it would make sense to list CU and L2 cacher per XCD for MI300 as well.


yhuiYH:
Thanks for the suggested fixes.  Since you added the XCD acronym, please add that to the Glossary at the bottom of the page as well (for completeness).
XCD = Accelerated Complex Die.

yhuiYH (approved):
Thanks, LGTM now.

PR #3947, opened 2024-10-25
streamlined the iommu section

spolifroni-amd:


docs/conceptual/iommu.rst
```diff
@@ -6,93 +6,51 @@
...
-  However, it should be noted that not using the IOMMU does not make peer to peer DMA just work.
-  There may be platform limitations as noted above that you could run into unless you know the limitations of your platform and configure it appropriately.
+The input-output memory management unit (IOMMU) provides memory remapping services for IO devices. 
+
+Similar to the memory management unit (MMU) on a CPU, the IOMMU is able to provide a private virtual address space for each device or group of devices connected to a CPU. 
+
+The IOMMU's memory remapping services are used in the following ways:
+
+* To provide private IO space for devices used in a guest virtual machine. 
+* To prevent unauthorized direct memory access (DMA) requests to the memory-managed IO (MMIO).
```
alexdeucher, line 16: MMIO = Memory Mapped IO.  Also, the IOMMU prevents unauthorized DMA access to system memory.  I would say something like:
To prevent unauthorized direct memory access (DMA) requests to system memory and memory mapped IO (MMIO).

docs/conceptual/iommu.rst
```diff
@@ -6,93 +6,51 @@
...
+* To help in debugging memory access issues.
+* To facilitate peer-to-peer DMA.
+
+The IOMMU also provides interrupt remapping, which is useful for devices that support multiple interrupts and for interrupt delivery on hardware platforms with several cores.
```
alexdeucher, line 20: s/several cores/large numbers of cores/

docs/conceptual/iommu.rst
```diff
@@ -6,93 +6,51 @@
...
+
+AMD CPUs prior to AMD Zen only supported peer-to-peer DMA for writes. On CPUs from AMD Zen and onwards, peer-to-peer DMA is fully supported. 
+
+To use peer-to-peer DMA on Linux, the following options must be enabled in your Linux kernel configuration:
+
+* ``CONFIG_PCI_P2PDMA``
+* ``CONFIG_DMABUF_MOVE_NOTIFY`` 
+* ``CONFIG_HSA_AMD_P2P``
+
+Devices connected via XGMI links, such as the MI50/60, MI100, MI200, and MI300 accelerators, have private XGMI links between GPUs and the PCIe bus is not used for peer-to-peer DMA. Because the PCIe bus is not used for peer-to-peer DMA, there are no device physical addressing limitations or platform root port limitations. However, because non-GPU devices such as RDMA NICs use PCIe for peer-to-peer DMA, there would be device physical addressing limitations and platform root port limitations when they interact with GPUs.
```
alexdeucher, line 56: s/there would be device physical/there may be device physical/
s/interact with GPUs/interact with other devices including GPUs/

alexdeucher: Finally, I think maybe we want a recommendations section.  I.e., now that we've given you the background, here's how it can come into play on your system.  E.g.,

For Instinct GPUs (MI200, MI300, etc.), AMD recommends setting iommu=pt.  This enables interrupt remapping which enables use with CPUs with large numbers of cores while keeping IO remapping disabled for the best compatibility with drivers from other vendors such as RDMA NICs.  Peer to Peer DMA between GPUs does not face any addressing limitations because MI GPUs have private XGMI links.
For Radeon GPUs, AMD also recommends iommu=pt.  However, if you need peer to peer DMA between GPUs it may be beneficial to enable the IOMMU to avoid any platform specific addressing limitations.

spolifroni-amd: I made changes around this and added recommendations in the settings table itself.

One question though: do Radeon GPUs not have XGMI links?

alexdeucher: Correct.  XGMI links are only available on MI parts.

PR #3958, opened 2024-10-30
Fix typo of prerequisites in ubuntu22 scripts

Wetitpig:
Rename `install-prerequisities.sh` to `install-prerequisites.sh`

prerequisit _i_ es seems like a typo of prerequisites and makes testing builds with different ubuntu OS versions inconvenient.


yhuiYH:
@WBobby or @amd-aakash please review as code-owners.  I don't have permission.

samjwu (approved):

PR #3959, opened 2024-10-30
Typo and venv in roc-6.2.x branch

Wetitpig:
The typo of `local-pin-60` instead of `local-pin-600` currently breaks the script in `install-prerequisites.sh`:
https://github.com/ROCm/ROCm/blob/e4ee91e83d4d8238cb17c871f255e2fe529f650b/tools/rocm-build/docker/ubuntu24/install-prerequisites.sh#L225

Moreover, simply initiailising a virtual environment with `python -m venv /opt/venv` does not activate it, causing system packages to still break in the following steps.

The virtual environment is hence activated right after creation.


yhuiYH:
@WBobby or @amd-aakash please help review these changes

PR #4137, opened 2024-12-07
Update default.xml

JeniferC99:
Remove ROCT-Thunck-Interface
Add TransferBench
Add hipBLAS-common

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -10,8 +10,8 @@
     <project name="ROCR-Runtime" />
     <project name="ROCT-Thunk-Interface" />
     <project name="amdsmi" />
-    <project name="omniperf" />
-    <project name="omnitrace" />
+    <project name="rocprofiler-compute" />
+    <project name="rocprofiler-systems" />
     <project name="rdc" />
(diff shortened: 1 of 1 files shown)


dgaliffiAMD (commented):
Do we need to add rocJPEG too? 

See: https://github.com/ROCm/ROCm/pull/4136

peterjunpark:
I think `<project name="ROCT-Thunk-Interface" />` should be removed

dgaliffiAMD:
> I think `<project name="ROCT-Thunk-Interface" />` should be removed

Thanks. I wasn't sure about that one.

dgaliffiAMD:
Let's re-open, we still have to remove "ROCT-Thunk-Interface" from the manifest.

dgaliffiAMD (approved):

dgaliffiAMD (dismissed):

JeniferC99:
I think we need to add hipBLAS-common and TransferBench too

peterjunpark (approved):

PR #4327, opened 2025-02-01
Adding TheRock to the manifest

ammallya:


powderluv (approved):
This should be NFC for regular users but gets TheRock collocated for ease of use as an alternate build system.

PR #4385, opened 2025-02-17
Docs: use custom directive to reference library versions

adeljo-amd:
The goal of this PR is to prevent margins of error where library versions across multiple files aren't updated correctly. A directive can be used in the page such as `:version-ref:``hipCUB latest`` or `:version-ref:``hipCUB 6.3.1``

### DIFF
--- a/docs/compatibility/ml-compatibility/jax-compatibility.rst
+++ b/docs/compatibility/ml-compatibility/jax-compatibility.rst
@@ -4,6 +4,8 @@
    :description: JAX compatibility
    :keywords: GPU, JAX compatibility
 
+.. version-set:: rocm_version latest
+
 *******************************************************************************
 JAX compatibility
 *******************************************************************************
@@ -119,7 +121,8 @@ Critical ROCm libraries for JAX
 
 The functionality of JAX with ROCm is determined by its underlying library
 dependencies. These critical ROCm components affect the capabilities,
-performance, and feature set available to developers.
+performance, and feature set available to developers. The versions described
+are available in ROCm :version:`rocm_version`.
 
 .. list-table::
     :header-rows: 1
@@ -129,7 +132,7 @@ performance, and feature set available to developers.
       - Purpose
       - Used in
     * - `hipBLAS <https://github.com/ROCm/hipBLAS>`_
-      - 2.3.0
+      - :version-ref:`hipBLAS rocm_version`
       - Provides GPU-accelerated Basic Linear Algebra Subprograms (BLAS) for
         matrix and vector operations.
       - Matrix multiplication in ``jax.numpy.matmul``, ``jax.lax.dot`` and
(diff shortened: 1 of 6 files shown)


docs/compatibility/ml-compatibility/jax-compatibility.rst
```diff
@@ -129,7 +130,7 @@ performance, and feature set available to developers.
       - Purpose
       - Used in
     * - `hipBLAS <https://github.com/ROCm/hipBLAS>`_
-      - 2.3.0
+      - :version-ref:`hipBLAS latest`
```
neon60, line 133: Can we store in one variable which version rocm version I would like to download for a specific framework?

adeljo-amd: Added, I tried RST substitution initially but that was a tedious approach. This approach works

neon60 (approved):

peterjunpark:
@alexxu-amd, do you want the custom directive in rocm-docs-core or is it fine here? I guess it's fine here because other projects probably won't use it?

peterjunpark (dismissed):

adeljo-amd:
> We need to move away from this unwieldy [CSV file](https://github.com/ROCm/ROCm/blob/develop/docs/compatibility/compatibility-matrix-historical-6.0.csv?plain=1) at some point IMO... We should use a more structured format, but I guess that will come later.

@peterjunpark Funny enough at one point I considered using yaml files, but after some discussion we decided to stick with the existing csv file. The script itself can be modified easily, but some manual effort would be needed to port the original csv to any new format.

What do you think? Would it be worthwhile to explore those changes in this PR?

peterjunpark:
> > We need to move away from this unwieldy [CSV file](https://github.com/ROCm/ROCm/blob/develop/docs/compatibility/compatibility-matrix-historical-6.0.csv?plain=1) at some point IMO... We should use a more structured format, but I guess that will come later.
> 
> @peterjunpark Funny enough at one point I considered using yaml files, but after some discussion we decided to stick with the existing csv file. The script itself can be modified easily, but some manual effort would be needed to port the original csv to any new format.
> 

@adeljo-amd I'm not sure what the other side of the discussion you had was, but perhaps we can just generate the CSV file if users have a need for that.

Manually maintaining the CSV file might be prone to human error. Also, it's not ideal to have `:ref:`s and other RST markup embedded in the [data](https://github.com/ROCm/ROCm/blob/develop/docs/compatibility/compatibility-matrix-historical-6.0.csv?plain=1) the way we currently do.

> What do you think? Would it be worthwhile to explore those changes in this PR?

I am working on improving [this compatibility table](https://rocm.docs.amd.com/en/latest/compatibility/compatibility-matrix.html) and I was planning on revisiting this there.

peterjunpark (dismissed):

adeljo-amd:
> @adeljo-amd I'm not sure what the other side of the discussion you had was, but perhaps we can just generate the CSV file if users have a need for that.
> 
> Manually maintaining the CSV file might be prone to human error. Also, it's not ideal to have `:ref:`s and other RST markup embedded in the [data](https://github.com/ROCm/ROCm/blob/develop/docs/compatibility/compatibility-matrix-historical-6.0.csv?plain=1) the way we currently do.

@peterjunpark Sure, just let me know how I can help when you revisit it. I'll just wait to see if Alex has any inputs before merging this PR in

peterjunpark (approved):

PR #4425, opened 2025-02-27
Update workload.rst with small export fix

dwiddows:
Tiny fix that removes the "export" directive. 
` export HIP_FORCE_DEV_KERNARG=1  hipblaslt-bench ...` leads to 
bash: export: `hipblaslt-bench': not a valid identifier

whereas just starting with HIP_FORCE_DEV_KERNARG=1 passes this env var to the hipblaslt-bench process, which I think is the intention here.


peterjunpark (approved):
LGTM

PR #4427, opened 2025-02-28
Debian build

cbayle:
Add build directory for debian12 and debian13
This is work in progress and require some extra changes you can see there:

https://github.com/ROCm/ROCm/compare/develop...cbayle:ROCm:build_changes_for_debian

Let me know if you want me to make some pull request about these extra changes


ThijsWithaar:
Ah, I had been looking at the same thing. Issues I ran into, which you might as well, are:

1. Only supports the "ubuntu" distro name: https://github.com/ROCm/ROCm/blob/develop/tools/rocm-build/build_aqlprofile.sh#L104
2. Trixie has gcc-14, which causes an error installing this deb: https://github.com/ROCm/ROCm/blob/develop/tools/rocm-build/build_lightning.sh#L515

zichguan-amd:
No longer relevant as we are moving forward with TheRock

Danny3:
Why doesn't AMD support Debian first?
Which means all derivatives and downstream projects like Ubuntu and Linux Mint will be supported too.
I don't see why not.

selroc:
> Why doesn't AMD support Debian first? Which means all derivatives and downstream projects like Ubuntu and Linux Mint will be supported too. I don't see why not.

I asked same question four years ago ...

zichguan-amd:
TheRock now builds off of `manylinux_2_28_x86_64` in an effort to support as many Linux distros as possible

Danny3:
> > Why doesn't AMD support Debian first? Which means all derivatives and downstream projects like Ubuntu and Linux Mint will be supported too. I don't see why not.
> 
> I asked same question four years ago ...

That's even more sad to hear!
Maybe there is something shady here, which I bet Canonical accepts easier, as they are already BFF with both Nvidia and Microsoft!

Debian on the other hand, being a community projects has much higher moral values, like the F-droid project!
Maybe that's why in both Debian and F-droid repositories (the main ones) is very hard to enter.
But once you are in, the users are guaranteed to have the best privacy and security standards respected and even reproducible builds for a lot of the packages / apps.
I really can't think of any other reason why a company that has enough money, hence manpower, is not will to support the upstream project and it's supporting just a derivative downstream distro with a not so good history and many times being called "The Windows of Linux".
Shame!

PR #4462, opened 2025-03-07
Ex CI: fixes for rocMLIR, rocPyDecode, RVS, rocprof-compute

danielsu-amd:
- rocMLIR
  - Install hip-python pip package, other python deps
  - For https://github.com/ROCm/rocMLIR/pull/1762
- rocPyDecode
  - Change hip-python installation method to match rocMLIR
- ROCmValidationSuite
  - Matrix-ize the test conf file directory to not be hardcoded for MI300X
  - Sample run: https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=22884&view=results
    - Run is based off https://github.com/ROCm/ROCmValidationSuite/pull/908 since it contains a critical build fix
- rocprofiler-compute
  - Removed a long-deprecated `ROCPROFCOMPUTE_ARCH_OVERRIDE` var that was being set for tests

### DIFF
--- a/.azuredevops/components/rocMLIR.yml
+++ b/.azuredevops/components/rocMLIR.yml
@@ -21,7 +21,7 @@ parameters:
     - numpy
     - ml_dtypes
     - scipy
-    - hip-python --index-url https://test.pypi/org/simple
+    - hip-python --extra-index-url https://test.pypi/org/simple
 - name: rocmDependencies
   type: object
   default:
--- a/.azuredevops/components/rocMLIR.yml
+++ b/.azuredevops/components/rocMLIR.yml
@@ -18,6 +18,10 @@ parameters:
   type: object
   default:
     - tomli
+    - numpy
+    - ml_dtypes
+    - scipy
+    - hip-python --index-url https://test.pypi/org/simple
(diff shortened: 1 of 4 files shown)


amd-jmacaran (approved):

.azuredevops/components/rocMLIR.yml
```diff
@@ -9,15 +9,19 @@ parameters:
...
     - libdrm-dev
     - libstdc++-12-dev
+    - ninja-build
+    - python3-pip
 - name: pipModules
   type: object
   default:
+    - hip-python --extra-index-url https://test.pypi.org/simple
+    - ml_dtypes
+    - numpy
```
umangyadav, line 22: Wouldn't it be better to actually run the same `requirements.txt`  before running the jobs ?

amd-jmacaran: As the requirements.txt is not standard fare for all the ROCm repos/components this CI system supports, we prefer to have better visibility within this same repo/project as to what components require what dependencies. It the future, we can transition to using such a file as the standard method.

PR #4491, opened 2025-03-13
Update vLLM performance Docker docs

peterjunpark:


docs/how-to/rocm-for-ai/inference/vllm-benchmark.rst
```diff
@@ -290,3 +314,40 @@ Further reading
...
+   * - ROCm version
+     - vLLM version
+     - PyTorch version
+     - Resources
+
+   * - 6.3.1
+     - 0.6.6
+     - 2.7.0
+     - 
+       * `Documentation <https://rocm.docs.amd.com/en/docs-6.3.2/how-to/rocm-for-ai/training/benchmark-docker/pytorch-training.html>`_
```
alexxu-amd, line 338: Overall looks good, a minor thing I noticed is the linked documentation version is different from the ROCm version stated. Is this intented?

peterjunpark: Yeah, this is intentional for now -- so we can point to the previous version of the vLLM performance docs. Definitely not the best solution though.

Eventually, we should decouple the vLLM Docker version from the ROCm version in the docs.

PR #4505, opened 2025-03-17
LLVM- project pointed to latest docs

prbasyal-amd:


randyh62 (approved):
Not the approach I would have taken, but looks ok to me.

PR #4525, opened 2025-03-24
Ex CI: dynamically set rocrtst include directory

danielsu-amd:
rocrtst's `OPENCL_INC_DIR` was being hardcoded for Clang 21, which was causing failures on mainline as mainline is still on Clang 19. 
Adds a script to dynamically get the latest available Clang directory and use that instead.

Staging build:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=24738&view=results

Mainline build:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=24737&view=results


darren-amd (approved):
LGTM

PR #4542, opened 2025-03-28
Ex CI: fix snap latest cmake version to 3.31

danielsu-amd:
CMake 4.0 is breaking a number of pipelines that use the `dependencies-cmake-latest` template, so fix the CMake version to `3.31/stable`.


darren-amd (approved):
LGTM

PR #4710, opened 2025-05-06
[Azure CI] [hipBLASLt] perl module dependencies added

jayhawk-commits:
- Tip of hipBLASLt has build failures when creating Tensile libraries. Iterative build attempts showed these two required modules were missing in default Azure VM.

### DIFF
--- a/.azuredevops/components/hipBLASLt.yml
+++ b/.azuredevops/components/hipBLASLt.yml
@@ -15,8 +15,10 @@ parameters:
   default:
     - git
     - libdrm-dev
+    - libfile-which-perl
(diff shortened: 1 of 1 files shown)


stellaraccident (approved):
This is fine, but note that the tools which were using these are deprecated and we are not adding new perl dependencies. I've already sent a revert: https://github.com/ROCm/hipBLASLt/pull/2028

If we land that revert, I kind of prefer that we keep CI deps minimal as it catches more stuff like this.

danielsu-amd:
The revert PR has been merged in, so I guess we don't need these perl deps anymore.

PR #4715, opened 2025-05-06
Ex CI: separate ROCgdb build and test jobs

danielsu-amd:
Separates the ROCgdb pipeline into a single build job and separate test jobs. 

Previously it was building+testing+uploading on both gfx942 and gfx90a systems, which led to two identical artifacts being uploaded. The ROCgdb build is GPU agnostic, so it only should upload one artifact.

Sample run:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=29019&view=results

### DIFF
--- a/.azuredevops/components/ROCgdb.yml
+++ b/.azuredevops/components/ROCgdb.yml
@@ -15,6 +15,7 @@ parameters:
   type: object
   default:
     - bison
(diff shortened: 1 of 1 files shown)


idass1990 (approved):
LGTM

PR #4734, opened 2025-05-13
Ex CI: remove firstRenderDeviceAccess demand from all components

danielsu-amd:
Removes the `firstRenderDeviceAccess` demand from all components that used it.

On our previous bare metal MI300X systems, any components that used `libva` would not work unless they had access to the "first" device on a system (`renderD128`). So we labelled the runner that had that device with the `firstRenderDeviceAccess` capability, and components were marked as requiring that capability to run.

In the current Kubernetes setup, this first device limitation no longer seems to be the case, so removing the demand.


idass1990 (approved):
LGTM

PR #4811, opened 2025-05-27
Ex CI: use clr amd-staging builds again

danielsu-amd:
amd-mainline clr is not functional. Instead we will switch back to amd-staging, delete all "bad" clr builds, and disable new builds until everything is stable again.

Currently I believe the bad builds started somewhere around https://github.com/ROCm/clr/commit/3cd3b3ffc5eba1def2b7a2b39ea9785dbdeddba7

### DIFF
--- a/.azuredevops/templates/steps/dependencies-rocm.yml
+++ b/.azuredevops/templates/steps/dependencies-rocm.yml
@@ -66,7 +66,7 @@ parameters:
       hasGpuTarget: false
     clr:
       pipelineId: $(CLR_PIPELINE_ID)
-      stagingBranch: amd-mainline
+      stagingBranch: amd-staging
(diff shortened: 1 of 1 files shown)


danielsu-amd:
rocPRIM last successful build was using https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=30620
Started failing with https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=30773

The https://github.com/ROCm/clr/commit/3cd3b3ffc5eba1def2b7a2b39ea9785dbdeddba7 commit is included in the diff between those two clr builds, so it's likely that 30620 is the last good build.

jayhawk-commits (approved):

PR #4849, opened 2025-05-30
[External CI] AlmaLinux8 job for llvm-project

jayhawk-commits:
- Removed building flang in this pipeline. Will build flang in the aomp pipeline to unblock progress on runtimes and first set of math libraries. Flang debug can also be moved to a cheaper VM.
- ninja-build from dnf is too old for llvm-project. Using a release from GitHub instead.
- Added more dnf package mappings.
- scl enable command is not needed.
- [Build Log](https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=31881&view=results)

### DIFF
--- a/.azuredevops/components/llvm-project.yml
+++ b/.azuredevops/components/llvm-project.yml
@@ -14,146 +14,168 @@ parameters:
   type: object
   default:
     - cmake
-    - python3-pip
     - libnuma-dev
     - ninja-build
-    - python-is-python3
-    - zlib1g-dev
     - pkg-config
+    - python-is-python3
+    - python3-pip
+    - zlib1g-dev
 - name: rocmDependencies
   type: object
   default:
     - rocm-cmake
 
+- name: jobMatrix
+  type: object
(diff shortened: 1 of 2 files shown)


.azuredevops/components/llvm-project.yml
```diff
@@ -14,146 +14,168 @@ parameters:
   type: object
   default:
     - cmake
-    - python3-pip
     - libnuma-dev
     - ninja-build
+    - pkg-config
     - python-is-python3
+    - python3-pip
```
danielsu-amd, line 21: Does python3-pip still need to be explicitly installed?

danielsu-amd (approved):

.azuredevops/components/llvm-project.yml
jayhawk-commits: It's surprisingly missing on the ultra VM. I think the Microsoft-hosted agent is the only ubuntu agent that has it preinstalled.

PR #4871, opened 2025-06-02
Ex CI: manifest changes to support monorepo & mathlibs builds

danielsu-amd:
Adds a sparse checkout directory field in manifests to differentiate between monorepo builds that originate from the same repo. Also adds in some logic to check if the current run is part of a unified mathlibs build.

Sample:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=32352&view=awardedsolutions.azure-pipelines-html-report-awardedsolutions.build-html-report-tab
image


idass1990 (approved):
LGTM

PR #4876, opened 2025-06-03
Ex CI: change rocprof-compute to use pip requirements.txt

danielsu-amd:
Changes the rocprofiler-compute pipeline to install pip packages from their requirements.txt files, which should allow PRs that add  new packages to successfully build in Azure CI. 

Also removed steps for downloading ROCm components during the build job, ROCm is only needed when running tests.

Moved the locale adding step to just before tests are run.

Sample run:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=32527&view=results

cc: @cfallows-amd

### DIFF
--- a/.azuredevops/components/rocprofiler-compute.yml
+++ b/.azuredevops/components/rocprofiler-compute.yml
@@ -22,26 +22,9 @@ parameters:
 - name: pipModules
   type: object
   default:
-    - astunparse==1.6.2
-    - colorlover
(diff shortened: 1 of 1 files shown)


jayhawk-commits:
Closing this PR. If we want to test if a dependency change is going to break something, we can make the test PR on the azure pipeline side. This gives us visibility on the change and result.

PR #4905, opened 2025-06-09
[Ex CI] add new rocprof-compute pip packages

danielsu-amd:
To fix build failures following the inclusion of new pip dependencies.

Removed steps for downloading ROCm components during the build job, ROCm is only needed when running tests.

Moved the locale adding step to just before tests are run.

Sample log (test failures are being looked at by team):
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=33529&view=results


jayhawk-commits (approved):
lgtm

PR #4918, opened 2025-06-12
[Ex CI] migrate roc/hipRAND pipelines, change migrated mathlibs's default branch to rocm-rel-7.0

danielsu-amd:
Changes the pipeline to pull artifacts from for roc/hipRAND to their monorepo pipelines.

For the mathlibs that are currently completely migrated over (PRIMs and RANDs), they are developing on their own branch: `release-staging/rocm-rel-7.0`. This means that changes related to HIP 7.0 are on that branch and not on `develop`.

To get these fixes into downstream component builds, change the staging branch for these components to that rel-7.0 branch, since its their staging branch in practice anyways.

### DIFF
--- a/.azuredevops/templates/steps/dependencies-rocm.yml
+++ b/.azuredevops/templates/steps/dependencies-rocm.yml
@@ -109,7 +109,7 @@ parameters:
       hasGpuTarget: false
     hipCUB:
       pipelineId: $(HIPCUB_PIPELINE_ID)
-      stagingBranch: develop
+      stagingBranch: release-staging/rocm-rel-7.0
       mainlineBranch: develop
       hasGpuTarget: true
     hipFFT:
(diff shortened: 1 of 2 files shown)


danielsu-amd:
Sample run to get MIOpen builds working:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=34020&view=results

The gfx90a test ran into a version mismatch with libamdhip.so, which is likely because MIOpen's upstream deps (hipBLASLt, rocRAND, etc.) have not been passing builds with HIP v7, so those haven't picked up the new so version yet. Once those components are fixed and are building again, I expect that MIOpen tests will be able to run after that.

hipBLASLt PR that should fix their builds: https://github.com/ROCm/hipBLASLt/pull/2207

The gfx942 test runner died mid-test, not yet sure if a one-off event or something that will need to be looked at.

jayhawk-commits (approved):
Let's try to keep this temporary and remember in the future. @idass1990

PR #4923, opened 2025-06-13
[Ex CI] fix hipRAND multi-OS tests, Tensile sparse dir

danielsu-amd:
Adds a missing job.os param to hipRAND test jobs.

Sample build:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=34294&view=results

Adds fixes for getting Tensile building from the monorepo.

Sample build:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=34300&view=results

### DIFF
--- a/.azuredevops/components/Tensile.yml
+++ b/.azuredevops/components/Tensile.yml
@@ -106,19 +106,19 @@ jobs:
       inputs:
         targetType: inline
         script: python3 setup.py bdist_wheel
-        workingDirectory: $(Build.SourcesDirectory)
+        workingDirectory: $(Agent.BuildDirectory)/s
     - task: Bash@3
       displayName: Rename wheel file with job OS
       inputs:
         targetType: inline
-        workingDirectory: $(Build.SourcesDirectory)
+        workingDirectory: $(Agent.BuildDirectory)/s
         script: |
-          wheelFile=$(find "$(Build.SourcesDirectory)/dist" -type f -name "*.whl" | head -n 1)
+          wheelFile=$(find "$(Agent.BuildDirectory)/s/dist" -type f -name "*.whl" | head -n 1)
           newWheelFile="$(basename "$wheelFile" .whl)-${{ job.os }}.whl"
           mv "$wheelFile" "$(dirname "$wheelFile")/$newWheelFile"
     - template: ${{ variables.CI_TEMPLATE_PATH }}/steps/artifact-prepare-package.yml
       parameters:
-        sourceDir: $(Build.SourcesDirectory)/dist
+        sourceDir: $(Agent.BuildDirectory)/s/dist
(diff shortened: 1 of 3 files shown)


.azuredevops/components/Tensile.yml
```diff
@@ -106,19 +106,19 @@ jobs:
...
         targetType: inline
         script: python3 setup.py bdist_wheel
-        workingDirectory: $(Build.SourcesDirectory)
+        workingDirectory: $(Agent.BuildDirectory)/s
     - task: Bash@3
       displayName: Rename wheel file with job OS
       inputs:
         targetType: inline
-        workingDirectory: $(Build.SourcesDirectory)
+        workingDirectory: $(Agent.BuildDirectory)/s
```
jayhawk-commits, line 114: Why is the switch to this path needed? What is wrong with Build.SourcesDirectory?

danielsu-amd: `Build.SourcesDirectory` is variable and changes to wherever the last repo checkout was, in the monorepo case it points to `$(Agent.BuildDirectory)/sparse` instead of `$(Agent.BuildDirectory)/s`. Everything else in the CI assumes that code is checked out under `/s`, so the hardcode is necessary.

The same path hardcode was put in `build-cmake` in https://github.com/ROCm/ROCm/pull/4701

https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services

jayhawk-commits: I suppose these references are unique to the Tensile build flow, because I was thinking maybe we make other component yml files match this string.

jayhawk-commits (approved):

.azuredevops/components/Tensile.yml
danielsu-amd: Normally when pulling from a dedicated repo, a component's source would be under `$(Build.SourcesDirectory)` (aka `$(Agent.BuildDirectory)/s`) and everything would work fine.

When pulling from the monorepo, the component source is now under `$(Build.SourcesDirectory)/projects/<component>` (aka `$(Agent.BuildDirectory)/sparse/projects/<component>`)

Instead of rewriting everything to handle the monorepo case, we're just symlinking the `$(Agent.BuildDirectory)/sparse/projects/<component>` dir into `$(Agent.BuildDirectory)/s`, which allows the existing CI to handle it without too much hassle.

danielsu-amd: This does mean we should change instances of `$(Build.SourcesDirectory)` to `$(Agent.BuildDirectory)/s`, but it's only a problem for the monorepo-enabled components and I've been changing those as I enable them.

PR #4959, opened 2025-06-23
[Ex CI] allow rerun jobs to upload artifacts

danielsu-amd:
Published artifacts in Azure are immutable and cannot be deleted or overwritten (unless the associated job is deleted as well). This caused issues when rerunning build jobs, as they'd try to publish artifacts with names that were already taken.

Adds `System.JobAttempt` at the end of artifact and file names, to differentiate between reruns and to allow artifacts to be published, as they will have unique names ([reference](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables)).

 Sample for rocminfo, was rerun once:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=35572&view=artifacts&pathAsName=false&type=publishedArtifacts

### DIFF
--- a/.azuredevops/components/rocminfo.yml
+++ b/.azuredevops/components/rocminfo.yml
@@ -84,7 +84,6 @@ jobs:
       parameters:
         os: ${{ job.os }}
     - template: ${{ variables.CI_TEMPLATE_PATH }}/steps/artifact-links.yml
-    - script: exit 1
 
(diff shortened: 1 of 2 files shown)


darren-amd (approved):
LGTM

PR #4986, opened 2025-06-30
[Ex CI] hipBLASLt: build some archs on medium pool

danielsu-amd:
For hipBLASLt, move everything except gfx942 onto the medium pool to save time and resources, as ultra pool is not currently equipped to handle its workload. For some reason they have the exact same build time as on the ultra pool, so no loss there.

Sample run:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=36597&view=results


darren-amd (approved):
LGTM

PR #4987, opened 2025-06-30
[Ex CI] migrate rocBLAS to monorepo

danielsu-amd:
- Change rocBLAS pipeline ID to monorepo pipeline
- Enable hipBLASLt -> rocBLAS downstream path


darren-amd (approved):
LGTM

PR #4996, opened 2025-07-02
Fix Docker run commands in Megatron-LM Docker doc

peterjunpark:


randyh62 (approved):
Looks good. Thanks for your effort.

PR #4998, opened 2025-07-02
[Ex CI] fix printed artifact download links

danielsu-amd:
- Fixed printed artifact links which were broken following https://github.com/ROCm/ROCm/pull/4959

Reverted the more-verbose artifact names in that PR, since it'd require a lot more params in `artifact-links.yml` for little benefit. Artifact names now solely depend on Azure-defined system-wide variables (indicated by the `$(Var.SomeName)` syntax), which do not require adding parameters into every single template (ie. `ARTIFACT_NAME="$(Agent.JobName)_$(System.JobAttempt)"`). This will still allow rerun build jobs to work properly.

- Separated manifest JSON creation/printing into different steps to help with visibility
- Saved the manifest filename into an exported `manifest_filename` var to avoid repeating the really long filename creation line

Sample run:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=37006&view=results

### DIFF
--- a/.azuredevops/templates/steps/artifact-links.yml
+++ b/.azuredevops/templates/steps/artifact-links.yml
@@ -15,8 +15,8 @@ steps:
       URL_BEGIN="https://artprodcus3.artifacts.visualstudio.com/"
       URL_MIDDLE="/_apis/artifact/"
       URL_END="/content?format=file&subPath=%2F"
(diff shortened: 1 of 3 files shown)


darren-amd (approved):
LGTM

PR #5005, opened 2025-07-07
[Ex CI] create Docker images for nightly builds

danielsu-amd:
Adds the creation of a Docker image to the nightly builds that:
- Installs a basic set of packages
  - Other packages will need to be installed manually, if needed
- Downloads and extracts aqlprofile
  - Can be removed once aqlprofile is added into CI
- Downloads and extracts the tarball from the previous steps
- Sets linker paths
- Adds `rocm/bin` to PATH

After the image is uploaded, it will print a single command to run the image.

Images are named: `rocmexternalcicd.azurecr.io/nightly-{os}-{gfx-target}-{source}:{build-id}`

Sample run:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=38531&view=results

Basic verification was performed for the gfx90a images on an MI250 system by running (UB24 only had the smi's available):
`rocminfo` `rocm_agent_enumerator` `rocm-smi` `amd-smi`

A potential consideration is that regularly pushing these nightly images to our registry may increase costs, since they're pretty large and this would happen on a daily basis.


jharryma (approved):
LGTM

PR #5122, opened 2025-07-30
[Ex CI] disable MIOpen downstream jobs

danielsu-amd:
MIOpen's downstream jobs (MIVisionX and MIGraphX) require them to be specified in a trigger file's `resources.repositories`, only the MIOpen pipeline currently has this. All other pipelines which include MIOpen as a downstream job are currently nonfunctional. 

Disable MIOpen downstream jobs until we come up with a solution.


jharryma (approved):

amd-hsivasun (approved):
LGTM

PR #5131, opened 2025-07-31
[Ex CI] MIOpen: use latest CK

danielsu-amd:
Change the MIOpen pipeline to pull latest CK instead of using the specific CK commit called out in MIOpen's requirements.txt.

Sample run:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=41431&view=results


danielsu-amd:
On hold while we figure out the future of the relationship between MIOpen and CK.

PR #5147, opened 2025-08-01
Update README.md removing reference to instructions that were moved

dwiddows:
Followup to the change in https://github.com/ROCm/ROCm/pull/4907/files - the instructions are no longer below.

(Could also move the previous paragraph into this section.)


peterjunpark (approved):
LGTM

PR #5148, opened 2025-08-01
Removing reference to previous instructions

dwiddows:
Slight extra to the change in https://github.com/ROCm/ROCm/pull/4907/files - removes a couple of lines that refer to instructions below that are no longer there.


peterjunpark:
Hi @dwiddows, is this the same as your other fix https://github.com/ROCm/ROCm/pull/5147?

dwiddows:
Yeah, looks like I managed to make 2 PRs from one commit? Will revert this one.

PR #5184, opened 2025-08-12
[Ex CI] add temporary downstream path from rocBLAS to hipBLAS

danielsu-amd:
Add in a temporary downstream path from rocBLAS to hipBLAS while the SOLVERs are still disabled, pending their migration.

Sample run for downstream functionality:
https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=43880&view=results


amd-hsivasun (approved):
LGTM

PR #5208, opened 2025-08-19
[Ex CI] change rocprofiler's branch to develop

danielsu-amd:
Change rocprofiler's branch name to `develop`, as rocm-systems does not use amd-staging.

This was overlooked because, for rocm-libraries, the mathlibs all already used the `develop` branch, so no change was needed for them. The components under rocm-systems have varying branch names, so it's now necessary to check this. 

Instructions on enabling monorepo support have been updated to reflect this new step.


amd-hsivasun (approved):
LGTM

PR #5838, opened 2026-01-07
[External CI] Add SIMDe dev package to HIP runtime pipeline

jayhawk-commits:
Latest clr/HIP looks for SIMDe on environment. Adding `libsimde-dev` in list of packages installed via apt.


amd-hsivasun (approved):
LGTM

PR #5965, opened 2026-02-13
[docs/7.11.0] fix `cd C:\TheRock`

peterjunpark:
## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/docs/install/includes/install.rst
+++ b/docs/install/includes/install.rst
@@ -982,7 +982,7 @@ Use the following instructions to install the ROCm Core SDK on your system.
 
    .. selected:: os=windows
 
-      Download the tarball and extract the contents to ``C:\TheRock``.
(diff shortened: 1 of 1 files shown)


docs/install/includes/install.rst
```diff
@@ -982,14 +982,14 @@ Use the following instructions to install the ROCm Core SDK on your system.
 
    .. selected:: os=windows
 
-      Download the tarball and extract the contents to ``C:\TheRock\build``.
```
lsudarsh-amd, line 985: While we download the tarball to "C:\TheRock", we are technically still extracting the contents to "C:\TheRock\build". I think its ok to use the old line of "Download the tarball and extract the contents to "C:\TheRock\build".

lsudarsh-amd (approved):

PR #6189, opened 2026-04-28
[docs/7.12.0] fix: prepend to LD_LIBRARY_PATH instead of clobbering

peterjunpark:
## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/docs/install/includes/post-install.rst
+++ b/docs/install/includes/post-install.rst
@@ -27,7 +27,7 @@ installation.
             .. code-block:: bash
 
                sudo tee /etc/profile.d/set-rocm-env.sh << EOF
-               export LD_LIBRARY_PATH=/opt/rocm/core/lib/rocm_sysdeps/lib:/opt/rocm/core/lib
+               export LD_LIBRARY_PATH=/opt/rocm/core/lib/rocm_sysdeps/lib:/opt/rocm/core/lib:\$LD_LIBRARY_PATH
                EOF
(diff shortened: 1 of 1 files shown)


docs/install/includes/post-install.rst
```diff
@@ -115,7 +115,7 @@ installation.
 
                      # BEGIN ROCm environment configuration
                      export ROCM_PATH=$ROCM_INSTALL_PATH
-                     export LD_LIBRARY_PATH=\$ROCM_PATH/lib/rocm_sysdeps/lib:\$ROCM_PATH/lib/
+                     export LD_LIBRARY_PATH=\$ROCM_PATH/lib/rocm_sysdeps/lib:\$ROCM_PATH/lib:\$LD_LIBRARY_PATH/
```
lsudarsh-amd: Small nitpick, we should standardize on always using a trailing / or not use it at all. I'd say let's leave it out for now.

docs/install/includes/post-install.rst
```diff
@@ -131,7 +131,7 @@ installation.
...
```
lsudarsh-amd: Small nitpick, we should standardize on always using a trailing / or not use it at all. I'd say let's leave it out for now.

lsudarsh-amd (changes requested):

lsudarsh-amd (approved):

PR #6366, opened 2026-06-16
Remove stored token logging from autotag script

fallintoplace:
## Summary
- Remove the log line that printed a stored GitHub token from the autotag script.

## Why
`get_token()` retrieved stored credentials and printed the full token value. That could expose a personal access token in CI logs or terminal scrollback.

## Validation
- `python3 -m py_compile tools/autotag/tag_script.py`
- `git diff --check`
- Confirmed the old `Using stored token` log string is no longer present.

### DIFF
--- a/tools/autotag/tag_script.py
+++ b/tools/autotag/tag_script.py
@@ -41,7 +41,6 @@ def get_token(
                     "No token was passed, and no stored token could be found."
                 )
             used_token = used_token_bytes.decode()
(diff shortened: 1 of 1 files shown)


peterjunpark (approved):
LGTM

As a sidenote, I'm not sure if anyone still uses this script @alexxu-amd @pmoutsias-amd. Maybe we can remove the autotag script entirely.

Issue #6544, opened 2026-07-29
[Build]: Avoid printing full Git configuration in profiler build logs


Aidenwu0209:
### Problem description

The Omnitrace and rocprofiler-systems build helpers print the complete local
`.git/config` before and after submodule synchronization.

The scripts already print `git submodule status` and
`git submodule status --cached`, which provide the relevant submodule
diagnostics. Printing the full configuration adds unrelated local and remote
configuration to build logs and can expose credential-bearing or internal
remote URLs when those are configured by a build environment.

### Affected revision

`964740985b6900a2401a8a1494285667d1e7e6aa`

### Affected files

- `tools/rocm-build/build_omnitrace.sh`
- `tools/rocm-build/build_rocprofiler-systems.sh`

### Expected behavior

The scripts should retain useful submodule status output without dumping the
complete Git configuration.

### Proposed fix

Remove the four `cat .git/config` calls. Do not change submodule initialization,
synchronization, update behavior, or existing status diagnostics.

harkgill-amd:
Hey @Aidenwu0209, the build scripts you're referring to here have been deprecated in favour of TheRock https://github.com/ROCm/TheRock. It's probably best to remove these to prevent any confusion https://github.com/ROCm/ROCm/pull/6572.

PR #6572, opened 2026-08-04
Remove deprecated tools/rocm-build folder

harkgill-amd:
## Motivation

Prevents any future confusion about recommended ROCm build path such as in https://github.com/ROCm/ROCm/issues/6544.

## Technical Details

`tools/rocm-build` has been deprecated post TheRock. Remove the build scripts that have been lingering in this repo.

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


peterjunpark (approved):

WBobby (commented):
Two things worth doing with the merge:
1. **Leave a tombstone.** Right now the folder just disappears. A short note in `tools/README.md` (or the top-level README) saying "use [TheRock](https://github.com/ROCm/TheRock) for current releases; for 7.2.x and earlier, check out the matching `rocm-X.Y.Z` tag" would save the next person from filing another issue when their bookmark 404s.
2. **Close what this obsoletes.** #6543 and #6545 both patch files this PR deletes, so they'll conflict. #6544 should get a closing note too — it's a leftover-debug cleanup rather than anything needing a backport, but if it stays open after the files are gone someone will re-file it.

Also worth explicitly closing #6544 along with this. It's a leftover-debug cleanup (two stray `cat .git/config` calls around the submodule sync), not something that needs a backport to the release tags — but if it just sits open after the files are gone, someone will eventually re-file it. A one-liner there saying the scripts were removed as deprecated would close the loop.

harkgill-amd:
> Leave a tombstone. Right now the folder just disappears. A short note in tools/README.md (or the top-level README) saying "use [TheRock](https://github.com/ROCm/TheRock) for current releases; for 7.2.x and earlier, check out the matching rocm-X.Y.Z tag" would save the next person from filing another issue when their bookmark 404s.

There aren't any references to these build scripts in our docs and we already point to TheRock in our top level [README](https://github.com/ROCm/ROCm/blob/develop/README.md), 
```
A new open-source build platform for ROCm is under development at https://github.com/ROCm/TheRock, featuring a unified CMake build with bundled dependencies, Windows support, and more.
```

> Close what this obsoletes. https://github.com/ROCm/ROCm/pull/6543 and https://github.com/ROCm/ROCm/pull/6545 both patch files this PR deletes, so they'll conflict. https://github.com/ROCm/ROCm/issues/6544 should get a closing note too — it's a leftover-debug cleanup rather than anything needing a backport, but if it stays open after the files are gone someone will re-file it.

Will close the issues/PRs referencing these scripts once this PR for their removal is merged.

WBobby:
Small correction on "there aren't any references to these build scripts in our docs" — true for `develop`, but not for our shipped docs.

Every docs branch from `docs/6.2.4` through `docs/7.2.4` still ships `docs/how-to/build-rocm.rst`, a "Build ROCm from source" page whose only content is a link to `https://github.com/ROCm/ROCm?tab=readme-ov-file#getting-and-building-rocm-from-source`. That link is unpinned, so it resolves against the default branch — where the anchor no longer exists. It's already dead today, e.g. https://rocm.docs.amd.com/en/docs-7.2.4/how-to/build-rocm.html.

Not a blocker for this PR (the breakage predates it), but it's why a tombstone is worth two lines: those pages are still live, and after this merge the scripts won't be findable on `develop` either. Something like "use TheRock; for 7.2.x and earlier, check out the matching `rocm-X.Y.Z` tag" would give those users somewhere to land.

harkgill-amd:
> Every docs branch from docs/6.2.4 through docs/7.2.4 still ships docs/how-to/build-rocm.rst, a "Build ROCm from source" page whose only content is a link to https://github.com/ROCm/ROCm?tab=readme-ov-file#getting-and-building-rocm-from-source. That link is unpinned, so it resolves against the default branch — where the anchor no longer exists. It's already dead today, e.g. https://rocm.docs.amd.com/en/docs-7.2.4/how-to/build-rocm.html.

Yeah that looks like it's been broken for a while. I think it makes the most sense to update those docs branches to point to TheRock instead of trying to patch up the README to support the broken links. We should be handling that in separate docs related PRs.

@peterjunpark, any idea how far back we'd want to go with fixing https://github.com/ROCm/ROCm?tab=readme-ov-file#getting-and-building-rocm-from-source in previous docs releases?

harkgill-amd:
@WBobby, are we good to go forward with this one? The docs changes can follow up in separate PRs as they'll target the historic docs branches.

